PR #1, opened 2020-02-19
Update README.md

jamesefhawkins:
Consolidating the docs. You can see this here instead:

https://github.com/PostHog/posthog/wiki/Ruby-integration


timgl:
Is it worth linking straight to the right page?

PR #2, opened 2020-12-14
Update default host

yakkomajuri:
Even though @fuziontech now made sure t.posthog.com works, I guess it makes sense to keep this consistent with other libs and use app.posthog.com?﻿

### DIFF
--- a/lib/posthog/defaults.rb
+++ b/lib/posthog/defaults.rb
@@ -1,7 +1,7 @@
 class PostHog
   module Defaults
     module Request
-      HOST = 't.posthog.com'
+      HOST = 'app.posthog.com'
       PORT = 443
       PATH = '/batch/'
       SSL = true


fuziontech:
I actually think we should keep t.posthog.com and move everything to that. We should have a separate subdomain for events.

yakkomajuri:
Ok yeah makes sense to me - however you prefer to structure it. This was just a matter of consistency since all other libs use app

timgl:
@fuziontech this is what we had before and it caused a ton of issues, especially with the toolbar and trying to authenticate across that.

mariusandra:
Also for plugins, slack webhooks, etc we basically need the `site_url` to refer to the right domain. Otherwise the slack webhook will contain a link to t.posthog.com/actions, etc.

fuziontech:
This all makes sense. Yeah, I'd pull the trigger on this then. Eventually we'll probably want to throw in our own routing layer for this on app.posthog

mariusandra:
Version `1.1.0` of posthog-ruby pushed out with the change!

From what I know of the ruby community though, nobody updates their dependencies. So we'll be stuck with this domain for a very long time.

PR #12, opened 2022-01-14
present? is not available outside of Rails

jamilbk:
Hey guys! Just a headsup -- `present?` is not available outside of Rails. This causes client creation to fail with `undefined method .present? for nil::nilClass`.

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -31,7 +31,7 @@ class PostHog
 
       check_api_key!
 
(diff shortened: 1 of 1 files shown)


timgl:
@mariusandra I think this looks fine?

mariusandra:
I think this will work, as `personal_api_key` will most likely never be `""` (`"".present? == false` and `!!"" == true`).

mariusandra:
Version 1.2.3 released with the change.

Issue #13, opened 2022-05-30
`create_alias` is called incorrectly


macobo:
This came from a client: ticket [here](https://posthoghelp.zendesk.com/agent/tickets/39)

We set distinct_id for `create_alias` events as explicitly as `nil` in our code, but we should set it to the (previous) distinct_id as e.g. in the python library: https://github.com/PostHog/posthog-python/blob/9d2fa72753ae9e5f4c8095f56f647205c5dac4f7/posthog/client.py#L203-L218

AxelTheGerman:
Thanks for getting that fix out so quickly!

PR #14, opened 2022-05-30
Fix posthog.alias call

macobo:
Fixes https://github.com/PostHog/posthog-ruby/issues/13

We were rejecting calls for this before as distinct_id has been mandatory for some time.


tiina303 (approved):
LGTM

PR #16, opened 2022-06-21
Multivariate flags support

liyiy:
multivariate support with the v2 endpoint


lib/posthog/feature_flags.rb
```diff
@@ -56,7 +55,7 @@ def is_feature_enabled(key, distinct_id, default_result = false)
       else
         data = { 'distinct_id' => distinct_id }
         res = _request('POST', 'decide', false, data)
```
neilkakkar, line 57: make this `v=2`, instead of doing it in `_request` ?

neilkakkar (commented):
I think we're just missing a `get_feature_flag` call, otherwise LGTM!

neilkakkar:
ah, and `send_feature_flags` param for the capture call, which should probs go in `FieldParser`

liyiy:
> ah, and `send_feature_flags` param for the capture call, which should probs go in `FieldParser`

yep yep! i'll do that in another PR so it's easier to review

lib/posthog/feature_flags.rb
liyiy: wym? `_request` turns it into `/decide/?v=2/?token` which is wrong :(  if I pass in `decide/?v=2` here

liyiy: Also do we really need the `get_feature_flag` call since `is_feature_enabled` essentially does the same thing? 😮

liyiy: it's grabbing the object value so it'll return the string of whatever variant it's at if it's multivariate for `is_feature_enabled`

liyiy: i mean though i guess technically "is_feature_enabled" would be incorrect if it's returning a string instead of a boolean... which means the python library needs fixing too... what do you think?

neilkakkar: Yep, python and this one both need `get_feature_flag`, as otherwise we have no way to know the variant in the response.

And like you say, `is_feature_enabled` is a boolean always.

neilkakkar: oh boy, that request implicit parsing is very annoying..

There's a `join` function to seamlessly merge params though: https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI/RFC2396_Parser.html#method-i-join-label-Args

Or better yet, you can change the `_request` signature to accept extra query params, which you can append at the end of the URI generation.

The way it is right now is sadly too confusing for whoever comes in the future :/

lib/posthog/feature_flags.rb
```diff
@@ -76,6 +75,21 @@ def load_feature_flags(force_reload = false)
       end
     end
 
+    def get_feature_variants(distinct_id)
+      request_data = {
+            "distinct_id": distinct_id,
+            "personal_api_key": @personal_api_key,
```
liyiy, line 81: are groups required for this?

neilkakkar: yep, groups support multivariate flags too. So the signature is exactly the same as `is_feature_enabled`

liyiy: OKKK I ADDED ITTT

lib/posthog/feature_flags.rb
```diff
@@ -92,9 +109,10 @@ def _load_feature_flags()
     end
 
     def _request(method, endpoint, use_personal_api_key = false, data = {})
-      uri = URI("https://#{@host}/#{endpoint}/?token=#{@project_api_key}")
+      uri = URI("https://#{@host}/#{endpoint}")
       req = nil
       if use_personal_api_key
+        uri = URI("https://#{@host}/#{endpoint}/?token=#{@project_api_key}")
```
neilkakkar, line 115: I think we always need a `project_api_key`, otherwise the call wouldn't work.

How we determine project_api_key token: https://github.com/PostHog/posthog/blob/master/posthog/api/utils.py#L104

So, your idea elsewhere of just `POST`ing the `api_key` is very sound. We should do it in both cases, and not put it in the request anymore

liyiy: I think we went over this and agreed this was fine for now?

neilkakkar: yep yep

neilkakkar (approved):

PR #25, opened 2023-01-19
Fix hash keys not a symbol for splat args test error

liyiy:
Some local eval tests were failing locally because double splat takes hash keys as symbols instead of as strings 🤔 

Not sure how they were passing in the CI for previous PRs

### DIFF
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -140,7 +140,10 @@ class PostHog
       if fallback_to_decide && !only_evaluate_locally
         begin
           flags = get_feature_variants(distinct_id, groups, person_properties, group_properties)
+          symbolize_keys! response
+          symbolize_keys! flags
           response = {**response, **flags}
+          response = stringify_keys(response || {})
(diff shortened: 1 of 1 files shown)


neilkakkar:
I think you're on a too old ruby version, what version are you on?

But yeah, if the version you're on is supported as well, makes sense to fix this

liyiy:
> 

2.6 i think? What version are you on? Does it not fail for you locally?

liyiy:
> I think you're on a too old ruby version, what version are you on?
> 
> But yeah, if the version you're on is supported as well, makes sense to fix this


2.6 i think? What version are you on? Does it not fail for you locally?

neilkakkar:
Ah, yeah, looks like 2.6 is ded: https://endoflife.date/ruby

CI runs on 2.7, I'm on 2.7.1, yep doesn't fail locally (I'd be a shady person if it failed and I somehow still managed to get it in prod 😂 )

liyiy:
coooool closing then

Issue #24, opened 2022-11-15
Gem version limitation on "concurrent-ruby"


jordan-brough:
PR https://github.com/PostHog/posthog-ruby/pull/19 added a minor-version constraint to concurrent-ruby, without any explanation that I can see:

https://github.com/PostHog/posthog-ruby/pull/19/files#diff-ed813f6c69cb6085908b418178cfcc6a65bc126ca0b613e73116ce7be2675c0a

```diff
-  spec.add_dependency 'concurrent-ruby'
+  spec.add_dependency "concurrent-ruby", "~> 1", "< 1.1.10"
```

Is the `< 1.1.10` intentional and needed?

neilkakkar:
There _was_ some warning spew in 1.1.10 around some deprecation that I don't remember anymore (should've commented this!)

Will re-check if it's still there, and remove the limitation if not, thanks for raising!

milani:
I don't know how to check for those warnings, but I didn't see anything when running tests.

liyiy:
Version 2.2.0 updated to remove this limitation! https://rubygems.org/gems/posthog-ruby/versions/2.2.0

liyiy:
Leaving it open for now for visibility in case others bump into this issue too.

milani:
Thanks @liyiy!

PR #26, opened 2023-01-19
Remove concurrent gem limitation

liyiy:
Remove concurrent gem limitation, it doesn't seem to have any deprecation warnings tied to it anymore?

https://github.com/PostHog/posthog-ruby/issues/24

### DIFF
--- a/posthog-ruby.gemspec
+++ b/posthog-ruby.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
   spec.license = 'MIT'
   spec.required_ruby_version = '>= 2.0'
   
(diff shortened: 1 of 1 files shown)


posthog-ruby.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
   spec.license = 'MIT'
   spec.required_ruby_version = '>= 2.0'
   
-  spec.add_dependency "concurrent-ruby", "~> 1", "< 1.1.10"
+  spec.add_dependency "concurrent-ruby"
```
neilkakkar, line 18: we should keep it pinned though! Otherwise this can break under us

neilkakkar (approved):
looks good!

PR #27, opened 2023-01-24
Feature flags v3 payloads

liyiy:
Implements https://github.com/PostHog/posthog/issues/13627 for posthog-ruby
- get_feature_flag_payload
- get_all_flags_and_payloads
- change api to v=3

### DIFF
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -25,6 +25,7 @@ class PostHog
       @feature_flags = Concurrent::Array.new
       @group_type_mapping = Concurrent::Hash.new
       @loaded_flags_successfully_once = Concurrent::AtomicBoolean.new
+      @feature_flags_by_key = nil
 
       @task =
         Concurrent::TimerTask.new(
@@ -118,17 +119,28 @@ class PostHog
     end
 
     def get_all_flags(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
-      # returns a string hash of all flags
+    # returns a string hash of all flags
+      flags = get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
+    end
 
-      # make sure they're loaded on first run
+    def get_all_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
+      _, payloads = get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
+      payloads
+    end
+
+    def get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
       load_feature_flags
 
       response = {}
+      payloads = {}
       fallback_to_decide = @feature_flags.empty?
 
       @feature_flags.each do |flag|
         begin
-          response[flag[:key]] = _compute_flag_locally(flag, distinct_id, groups, person_properties, group_properties)
+          match_value = _compute_flag_locally(flag, distinct_id, groups, person_properties, group_properties)
+          response[flag[:key]] = match_value
+          match_payload = _compute_flag_payload_locally(flag, match_value)
+          payloads[flag[:key]] = match_payload
         rescue InconclusiveMatchError => e
           fallback_to_decide = true
         rescue StandardError => e
@@ -145,7 +157,43 @@ class PostHog
           logger.error "Error computing flag remotely: #{e}"
         end
       end
+      response, payloads, fallback_to_decide
+    end
+
+    def get_feature_flag_payload(key, distinct_id, match_value = nil, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
+      if match_value == nil
+        match_value = get_feature_flag(
+          key,
+          distinct_id,
+          groups=groups,
+          person_properties=person_properties,
+          group_properties=group_properties,
+          only_evaluate_locally=true,
+        )
+      end
+      response = nil
+      if match_value != nil
+        response = compute_payload_locally(key, match_value)
+      end
+      if response != nil and !only_evaluate_locally
+        decide_payloads = get_all_payloads(distinct_id, groups, person_properties, group_properties)
+        response = decide_payloads[key.downcase.to_sym] || nil
+      end
       response
+    end
+
+    def compute_payload_locally(key, match_value):
+      payload = nil
+
+      if @feature_flags_by_key == nil
+          return payload
+
+      flag_definition = @feature_flags_by_key[key] || {}
+      flag_filters = flag_definition["filters"] || {}
+      flag_payloads = flag_filters["payloads"] || {}
+      payload = flag_payloads[match_value.to_str.downcase]  || nil
+      
+      payload
     end
 
     def shutdown_poller()
@@ -248,6 +296,17 @@ class PostHog
 
     end
 
+    def _compute_flag_payload_locally(key, match_value)
+      response = nil
+      if [true, false].include? match_value
+        response = @feature_flags_by_key.dig(key.to_sym, :filters, :payloads, match_value.to_str.to_sym)
+      elsif match_value.is_a? String
+        response = @feature_flags_by_key.dig(key.to_sym, :filters, :payloads, match_value.to_sym)
+      end
+
+      response
+    end
+
     def match_feature_flag_properties(flag, distinct_id, properties)
       flag_filters = flag[:filters] || {}
 
@@ -346,7 +405,13 @@ class PostHog
       if !res.key?(:flags)
         logger.error "Failed to load feature flags: #{res}"
       else
-        @feature_flags = res[:flags] || []        
+        @feature_flags = res[:flags] || []
+        @feature_flags_by_key = {}
+        @feature_flags.each do |flag|
+          if flag[:key] != nil
+            @feature_flags_by_key[flag[:key]] = flag
+          end
+        end
         @group_type_mapping = res[:group_type_mapping] || {}
 
         logger.debug "Loaded #{@feature_flags.length} feature flags"
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -162,8 +162,8 @@ class PostHog
           flags_and_payloads = get_decide(distinct_id, groups, person_properties, group_properties)
           decide_flags = stringify_keys(flags_and_payloads[:featureFlags] || {})
           decide_payloads = stringify_keys(flags_and_payloads[:featureFlagPayloads] || {})
-          flags = {**flags, **decide_flags}
-          payloads = {**payloads, **decide_payloads}
+          flags = decide_flags
+          payloads = decide_payloads
         rescue StandardError => e
           logger.error "Error computing flag remotely: #{e}"
         end
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -120,12 +120,12 @@ class PostHog
 
     def get_all_flags(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
     # returns a string hash of all flags
-      response = get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally)
+      response = get_all_flags_and_payloads(distinct_id, groups, person_properties, group_properties, only_evaluate_locally)
       flags = response[0]
     end
 
     def get_all_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
-      response = get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally)
+      response = get_all_flags_and_payloads(distinct_id, groups, person_properties, group_properties, only_evaluate_locally)
       payloads = response[1]
     end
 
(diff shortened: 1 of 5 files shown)


lib/posthog/feature_flags.rb
```diff
@@ -137,13 +150,35 @@ def get_all_flags(distinct_id, groups = {}, person_properties = {}, group_proper
         end
       end
 
-      if fallback_to_decide && !only_evaluate_locally
```
liyiy, line 140: Node library says we keep this but python one says we remove this.... not sure what's supposed to happen here :|

lib/posthog/feature_flags.rb
```diff
@@ -118,33 +131,71 @@ def get_feature_flag(key, distinct_id, groups = {}, person_properties = {}, grou
...
 
     def get_all_flags(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
-      # returns a string hash of all flags
+    # returns a string hash of all flags
+      response = get_all_flags_and_payloads(distinct_id, groups, person_properties, group_properties, only_evaluate_locally)
+      flags = response[:featureFlags]
+    end
 
-      # make sure they're loaded on first run
+    def get_feature_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
```
neilkakkar, line 139: we don't need this function

lib/posthog/feature_flags.rb
```diff
@@ -118,33 +131,71 @@ def get_feature_flag(key, distinct_id, groups = {}, person_properties = {}, grou
...
           fallback_to_decide = true
         end
       end
-
       if fallback_to_decide && !only_evaluate_locally
         begin
-          flags = get_feature_variants(distinct_id, groups, person_properties, group_properties)
-          response = {**response, **flags}
+          flags_and_payloads = get_decide(distinct_id, groups, person_properties, group_properties)
+          symbolize_keys! flags
```
neilkakkar, line 169: this seems unnecessary?

neilkakkar: (this = symbolizing and then stringifying)

spec/posthog/feature_flag_spec.rb
```diff
@@ -3322,6 +3324,346 @@ class PostHog
...
+      ).to_return(status: 200, body: {"flags": [basic_flag_res]}.to_json)
+      stub_request(:post, decide_endpoint)
+        .to_return(status: 200, body:{"featureFlagPayloads": {"person-flag" => 300}}.to_json)
+      c = Client.new(api_key: API_KEY, personal_api_key: API_KEY, test_mode: true)
+      expect(c.get_feature_flag_payload("person-flag", "some-distinct-id", person_properties: {"region": "USA"})).to eq(300)
+      expect(c.get_feature_flag_payload("person-flag", "some-distinct-id", match_value: true, person_properties: {"region": "USA"})).to eq(300)
+      # assert_requested :post, decide_endpoint, times: 2
+    end
+
+    it 'with multivariate feature flag payloads' do
```
neilkakkar, line 3623: we need a test with a json payload coming from decide, to ensure it's parsed properly

spec/posthog/feature_flag_spec.rb
```diff
@@ -3322,6 +3324,346 @@ class PostHog
...
+      stub_request(
+        :get,
+        'https://app.posthog.com/api/feature_flag/local_evaluation?token=testsecret'
+      ).to_return(status: 200, body: {"flags": [basic_flag_res]}.to_json)
+      stub_request(:post, decide_endpoint)
+        .to_return(status: 200, body:{"featureFlagPayloads": {"person-flag" => 300}}.to_json)
+      c = Client.new(api_key: API_KEY, personal_api_key: API_KEY, test_mode: true)
+      expect(c.get_feature_flag_payload("person-flag", "some-distinct-id", person_properties: {"region": "USA"})).to eq(300)
+      expect(c.get_feature_flag_payload("person-flag", "some-distinct-id", match_value: true, person_properties: {"region": "USA"})).to eq(300)
+      # assert_requested :post, decide_endpoint, times: 2
```
neilkakkar, line 3620: Was this not being called 2x?

neilkakkar: Ah, match value exists in the second request, should be one times?

neilkakkar: .. and person property exists, should be 0 times :P

lib/posthog/feature_flags.rb
neilkakkar: ah, I see how it's used.

Here's a proposal for cleaner code:

Just like we did for all_flags_and_payloads, create a `get_feature_flag_and_payload` function, which has all the logic for `get_feature_flag`, and returns the flag value, payload value, and if it was evaluated locally.

Then, use this function for both `get_feature_flag`, and `get_feature_flag_payload`

The crux here is that everything comes from decide anyway, so it doesn't make sense to artificially split these out, and then have to toggle only_evaluate_locally to first see if it was locally available and then make the decide call later, when `get_feature_flag` already handles this logic.

neilkakkar (changes requested):
getting there! 🚀 , few comments

lib/posthog/feature_flags.rb
liyiy: It doesn't merge properly though if they're not all one type (either symbols or strings). I can convert them to all strings instead of symbolizing? 🤷‍♀️

neilkakkar: what's a string and what's not? This should be uniform 🤔  - anyway seems like a potential bug, would be nice to get to the bottom. I think we deal with everything being symbols, iirc, so everything coming should be a symbol.

liyiy: Actually you were already stringifying the keys here in `get_feature_variants` with the flags that come back from the decide endpoint 😮  

As we're setting the `flags` and `payloads` hash above, we have string key value pairs because the keys come from the feature flag `key` value which is a string, but the decide endpoint flags will return symbolized flags. It looks like you had made a decision previously to convert the flag results from the decide endpoint from symbols to strings instead so I'll just stick with that and then this avoids having to change all the test specs from strings to symbols if it's not needed

lib/posthog/feature_flags.rb
liyiy: yep looks like I had a duplicate method named the same.. removed it 👍

spec/posthog/feature_flag_spec.rb
liyiy: figured out why and fixed it. Was calling the wrong method somewhere and also passing in python syntax arguments instead of ruby somewhere else 😅

spec/posthog/feature_flag_spec.rb
liyiy: @neilkakkar ^ ?

neilkakkar: Ah, I meant that all payloads are string/int, but I don't see an array or json payload.

Just updating that test above to have real json payloads would work :)

spec/posthog/feature_flag_spec.rb
neilkakkar, line 3620: this comment, needs to be uncommented?, and times = 0?

lib/posthog/feature_flags.rb
```diff
@@ -49,7 +50,24 @@ def load_feature_flags(force_reload = false)
...
+        raise DecideAPIError.new(decide_data.to_json)
+      else
+        stringify_keys(decide_data[:featureFlags] || {})
+      end
+    end
+
+    def get_feature_payloads(distinct_id, groups = {}, person_properties = {}, group_properties = {}, only_evaluate_locally = false)
+      decide_data = get_decide(distinct_id, groups, person_properties, group_properties)
+      if !decide_data.key?(:featureFlagPayloads)
+        raise DecideAPIError.new(decide_data.to_json)
```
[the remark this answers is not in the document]
neilkakkar: as in, okay to raise here, but catch in call site

spec/posthog/feature_flag_spec.rb
liyiy: nah it should be 2, just fixed the test

lib/posthog/feature_flags.rb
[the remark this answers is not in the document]
neilkakkar: double check what happens when we just log here, as that means this function returns nil, not sure if we have existing  tests for this case 👀  & if it works as expected

lib/posthog/feature_flags.rb
```diff
@@ -346,7 +403,13 @@ def _load_feature_flags()
       if !res.key?(:flags)
         logger.error "Failed to load feature flags: #{res}"
       else
-        @feature_flags = res[:flags] || []        
+        @feature_flags = res[:flags] || []
```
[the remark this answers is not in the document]
liyiy: Isn't that fine on successful feature flags load? 

Also wouldn't this be broken on posthog python then too?? 😱 https://github.com/PostHog/posthog-python/blob/a9d7bf3e0b8bdd9c8b1037ad95b106d4cce15084/posthog/client.py#L378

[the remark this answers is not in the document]
neilkakkar: nope, that's precisely the problem, we lose flags in between flag loads.

It's not a problem in python because `get()` raises an error and doesn't update existing flags.

This comment was for Line 401, not the current line, but since it's too far away I can't comment on that line :P

lib/posthog/feature_flags.rb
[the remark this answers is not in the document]
neilkakkar: it will become a problem indeed for `get_feature_variants`, where we'll end up returning `nil` for get_all_flags, which shouldn't be possible (it's always a dict, either empty, or with things)

lib/posthog/feature_flags.rb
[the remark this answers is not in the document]
liyiy: But if the decide returns something without the flags key (an error), it's handled with the logger so it doesn't update existing flags right?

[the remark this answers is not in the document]
neilkakkar: unless I'm misreading something, line 401 clears flags, so if local evaluation errors out, we have empty flags, instead of the previous flags, which is not great.

Instead, we should've kept the old flags.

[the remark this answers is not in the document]
liyiy: Are you talking about this line? `@feature_flags.clear`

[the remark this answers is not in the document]
neilkakkar: yes, the line 401 :P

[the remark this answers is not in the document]
liyiy: Screen Shot 2023-02-06 at 2 41 28 PM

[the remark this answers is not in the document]
liyiy: ban implicit statements 😤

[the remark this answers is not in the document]
neilkakkar: oh noooo 😂 , in my defence, it's line 401 in the PR, which is all I have access to, not your ide lol.

Smh

lib/posthog/feature_flags.rb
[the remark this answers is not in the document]
liyiy: I'm returning an empty hash on missing key error, not 100% sure how to test this 👀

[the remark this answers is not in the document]
liyiy: Actually looks like we're expecting a `nil` return from errors so it should be fine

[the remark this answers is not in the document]
neilkakkar: cool

neilkakkar (approved):
Nice work!

I'd hope to see resiliency test in a follow up PR! 😅

lib/posthog/feature_flags.rb
```diff
@@ -119,32 +128,65 @@ def get_feature_flag(key, distinct_id, groups = {}, person_properties = {}, grou
...
           fallback_to_decide = true
         end
       end
-
       if fallback_to_decide && !only_evaluate_locally
         begin
-          flags = get_feature_variants(distinct_id, groups, person_properties, group_properties)
-          response = {**response, **flags}
+          flags_and_payloads = get_decide(distinct_id, groups, person_properties, group_properties)
+          decide_flags = stringify_keys(flags_and_payloads[:featureFlags] || {})
```
neilkakkar, line 160: can simplify :)

```suggestion
          flags = stringify_keys(flags_and_payloads[:featureFlags] || {})
```

neilkakkar: and same below

liyiy: yesssirr

liyiy:
> Nice work!
> 
> I'd hope to see resiliency test in a follow up PR! 😅

I'm not sure what resiliency test you mean other than one for the local evaluation `@feature_flags.clear` fix that didn't have a test for it to begin with 😅 but I added a test for that so if it looks good to you, feel free to merge this! Otherwise let me know what you meant and I'll address it in the morning 👍

neilkakkar:
that's exactly it, lovely :)

spec/posthog/feature_flag_spec.rb
```diff
@@ -3322,6 +3324,375 @@ class PostHog
...
+
+      stub_request(
+        :get,
+        'https://app.posthog.com/api/feature_flag/local_evaluation?token=testsecret'
+      ).to_return(status: 400, body: {"error": "went_wrong!"}.to_json)
+
+      # force reload to simulate poll interval
+      c.reload_feature_flags
+
+      expect(c.get_feature_flag("person-flag", "distinct_id", person_properties: {"region" => "USA"})).to eq(true)
```
neilkakkar, line 3695:
```suggestion
      expect(c.get_feature_flag("person-flag", "distinct_id", person_properties: {"region" => "USA"})).to eq(true)
      assert_not_requested :post, decide_endpoint
```

PR #34, opened 2023-11-23
feat(flags): add relative date operators and fix numeric ops

liyiy:
ruby version of the relative date operator add and numeric ops fixes for local evaluation

https://github.com/PostHog/posthog-python/pull/105

### DIFF
--- a/spec/posthog/feature_flag_spec.rb
+++ b/spec/posthog/feature_flag_spec.rb
@@ -1336,41 +1336,6 @@ class PostHog
       end
     end
 
-    # def test_none_property_value_with_all_operators(self):
-    #   property_a = self.property(key="key", value="none", operator="is_not")
-    #   self.assertFalse(match_property(property_a, {"key": None}))
-    #   self.assertTrue(match_property(property_a, {"key": "non"}))
-
-    #   property_b = self.property(key="key", value=None, operator="is_set")
-    #   self.assertTrue(match_property(property_b, {"key": None}))
-
-    #   property_c = self.property(key="key", value="no", operator="icontains")
-    #   self.assertTrue(match_property(property_c, {"key": None}))
-    #   self.assertFalse(match_property(property_c, {"key": "smh"}))
-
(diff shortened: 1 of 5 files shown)


spec/posthog/feature_flag_spec.rb
```diff
@@ -1220,6 +1234,121 @@ class PostHog
...
+        expect(FeatureFlagsPoller.match_property(property_n, { 'key' => '2022-05-01 00:00:00' })).to be true
+        expect(FeatureFlagsPoller.match_property(property_n, { 'key' => '2021-05-01 00:00:01' })).to be true
+        expect(FeatureFlagsPoller.match_property(property_n, { 'key' => '2021-05-01 00:00:00' })).to be false
+        expect(FeatureFlagsPoller.match_property(property_n, { 'key' => '2021-04-30 00:00:00' })).to be false
+        expect(FeatureFlagsPoller.match_property(property_n, { 'key' => '2021-03-01 12:13:00' })).to be false
+
+      end
+    end
+
+    it 'nil property value with all operators' do
```
liyiy, line 1339: Do we need to test this fully for ruby? Not sure it makes sense here since nil just returns empty when it's stringified. Unless we want to force nil values to be a string of nil.

neilkakkar: nope, makes sense, just make sure that operations don't crash with nil where we don't expect them to

spec/posthog/feature_flag_spec.rb
```diff
@@ -1336,19 +1336,169 @@ class PostHog
       end
     end
 
-    it 'nil property value with all operators' do
+    # def test_none_property_value_with_all_operators(self):
```
liyiy, line 1339: some leftover copy and paste (ignore if you're still WIPing it)

neilkakkar: nah this one's ready, will remove thanks!

PR #40, opened 2024-02-07
Add newer ruby versions to GH actions

jclusso:


jclusso:
Unfortunately it seems GH actions won't run without an approval so I don't know if this works for certain.

jclusso:
@neilkakkar I don't think we will be able to get this to pass until we merge #41.

jclusso:
Alright, rebased this and it looks like that did the trick @neilkakkar.

PR #41, opened 2024-02-07
Upgrade development dependencies

jclusso:
All of the dev dependencies were super outdated and the tests wouldn't work.

PR #42, opened 2024-02-07
Fix rspec output and warnings

jclusso:
**Note**: This requires #41 and can be rebased once that is merged.

The tests were previously outputing a ton of stuff from the logger. I've silenced the logger. Additionally, I've fixed deprecation warnings.


neilkakkar:
Nice 👍

Issue #10, opened 2021-09-12
Posthog ruby does not work in background job systems like Sidekiq and Resque


ajsharp:
## Problem

The posthog ruby client uses a background thread to send requests to posthog. This doesn't work in Resque or background job systems, because Resque doesn't wait for background threads to complete before completing a job. 

I see [this line](https://github.com/PostHog/posthog-ruby/blob/fad181e2106d755e61a1e4bdb8139930b976acf2/lib/posthog/transport.rb#L43) logged, but the request is never finished.

Currently, the `Client#flush` method exists, but it has an arbitrary 0.1 second `sleep` call in it -- not ideal for a non-exit scenario.

## Temporary Solution

To get posthog to send events, I had to monkeypatch `PostHog::Client` to run the worker in the current thread, and `PostHog::Worker` to not run in a continuous loop, which would block the main thread.

This is my local monkey patched code:

```ruby
class PostHog::Client
  def ensure_worker_running
    @worker_mutex.synchronize { @worker.run_sync }
  end
end

class PostHog::Worker
  def run_sync
    return if @queue.empty?

    @lock.synchronize do
      consume_message_from_queue! until @batch.full? || @queue.empty?
    end

    res = @transport.send @api_key, @batch
    @on_error.call(res.status, res.error) unless res.status == 200

    @lock.synchronize { @batch.clear }
  end
end
```

## Suggested Solution

Ideally, you could configure the posthog client to do some version of the above at runtime. This way you could allow Posthog to run async if it's running in an app server context, and synchronously in a background worker context. Or, even better would be direct support for Resque or Sidekiq. There are other libraries that do something similar.

wuputah:
Interesting. I did not expect the client to be sending work into a background thread for me. I used ActiveJob reflexively to move the work into the background, which worked fine as just using the thread pool back-end at the moment. An async/sync option would be preferred, and please add this to the docs.

itadventurer:
It seems like the `PostHog::Worker` was renamed to `PostHog::SendWorker` in one of the previous releases. So, here is the updated Monkeypatch:

```ruby

class PostHog::Client
  def ensure_worker_running
    @worker_mutex.synchronize { @worker.run_sync }
  end
end

class PostHog::SendWorker
  def run_sync
    return if @queue.empty?

    @lock.synchronize do
      consume_message_from_queue! until @batch.full? || @queue.empty?
    end

    res = @transport.send @api_key, @batch
[...]

PR #43, opened 2024-02-24
Fix memory leak in PostHog::Client.new

lsegal:
The PostHog::Client calls Kernel#at_exit at each instantiation, which registers a process-global callback for each Client instance. This global callback is represented by a T_FILE entry in ObjectSpace.count_objects and is never GC'd for the life of the process.

Any process that instantiates multiple clients will be susceptible to this memory leak, which is very common in almost any long-running process, but in particular, the norm for Sidekiq / Resque environments. 

The good news is this line is completely unnecessary, since in the current implementation, the worker thread does not block the process from exiting, which means that setting this thread variable is unlikely to actually cause the thread to "gracefully exit" unless the code also blocked on Thread execution via Thread#join. In other words, because Thread#join is not used, this line is a no-op. This is all mostly irrelevant anyway, since there is no actual cleanup that is not already being performed by the basic exiting of a process, and an at_exit callback (or graceful exit logic) is simply not needed for HTTP clients.

To reproduce the leak:

```ruby
GC.disable
p ObjectSpace.count_objects[:T_FILE]
10.times { PostHog::Client.new; GC.start }
p ObjectSpace.count_objects[:T_FILE]
```

This fix will likely improve use cases for users in #10.

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -52,8 +52,6 @@ class PostHog
         )
       
       @distinct_id_has_sent_flag_calls = SizeLimitedHash.new(Defaults::MAX_HASH_SIZE) { |hash, key| hash[key] = Array.new }
-
-      at_exit { @worker_thread && @worker_thread[:should_exit] = true }
     end
 
     # Synchronously waits until the worker has cleared the queue.


neilkakkar:
Thanks @lsegal! Getting someone to review this shortly 👍

neilkakkar:
Will publish a new version shortly 👍

ibrahima:
> Any process that instantiates multiple clients will be susceptible to this memory leak, which is very common in almost any long-running process, but in particular, the norm for Sidekiq / Resque environments.

Curious about this point - for e.g. a web server process, is it expected to initialize one client and share it rather than instantiating one in each request that needs to communicate with PostHog? It does seem like instantiating a PostHog client is a bit expensive so it might make sense to do it once at server initialization time... though I dunno if you'd have to do it in a thread-safe way and if there's any state on a PostHog client that would be undesirable to persist across requests.

lsegal:
Nothing is stopping you from instantiating a single object if you plan on using this from your webapp server itself. I can't personally speak to how qualified the "a bit expensive" statement is, but if it is expensive, you can indeed share the instance, it already was thread-safe before this patch, and this patch doesn't change anything about it's thread safety.

ibrahima:
Yeah that's fair, with the memory leak fixed the impact of making a new client is less severe, and it might already not be noticeable - I just noticed it because I'm working on a newish app and was profiling memory usage. That's good to know that it was thread safe already!

ibrahima:
Hmm, did a little more digging into this because I happened to notice that there was nothing that makes a thread set `:should_exit` now, so threads should never exit on their own https://github.com/PostHog/posthog-ruby/blob/7a92241d718e90b11dee22e25e6ecb91db6fa839/lib/posthog/send_worker.rb#L37.

It looks like in analytics-ruby, the gem on which this was based, this was actually added to avoid JRuby memory leaks. Kinda funny that it ended up causing a memory leak in other usage patterns.

https://github.com/segmentio/analytics-ruby/pull/28

For segment they do recommend initializing it as a constant in a Rails initializer: https://segment.com/docs/connections/sources/catalog/libraries/server/ruby/#getting-started. And it makes sense to me since the whole Queue/Thread pattern seems designed for long-running clients rather than spinning up a client on each request. In some somewhat anecdotal testing, it _feels_ like my application is a few ms faster on average when I initialize it in an initializer rather than on each request, which makes sense since it's doing less work per request. So I think this is the way to go for a Rails app! And maybe for a background worker too if it can be initialized once and shared.

PR #44, opened 2024-03-13
fix(flags): Add appropriate timeouts, fix some other misc things

neilkakkar:
This PR does a few things:

1. Add a `feature_flag_request_timeout_seconds` configuration option, defaulting to 3s to control flag timeouts
2. Gets rid of a lot of error spew, instead allows passing in an on_error proc so users can decide what to do with errors
3. Unifies calling sites such that everything doesn't call the `/decide` endpoint directly, but via `get_all_flags_and_payloads` which in turn does all necessary error handling. Before this, getting payloads would always throw :/
4. Noticed as a result of above^ that we were sometimes inefficiently going to decide. The test updates where I add a second feature flag key for `beta-feature2` reflect that, since previously this would go to decide, but now it wouldn't, which is infact the ideal thing to do. (and thus to test the old flow we need an extra flag that is present but can't be locally evaluated).

Noticed while doing this that the above problem exist in posthog-python as well, will update that too

### DIFF
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -162,6 +162,10 @@ class PostHog
       if fallback_to_decide && !only_evaluate_locally
         begin
           flags_and_payloads = get_decide(distinct_id, groups, person_properties, group_properties)
+
+          if !flags_and_payloads.key?(:featureFlags)
+            raise StandardError.new("Error flags response: #{flags_and_payloads}")
+          end
           flags = stringify_keys(flags_and_payloads[:featureFlags] || {})
           payloads = stringify_keys(flags_and_payloads[:featureFlagPayloads] || {})
         rescue StandardError => e
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -14,7 +14,7 @@ class PostHog
     include PostHog::Logging
     include PostHog::Utils
 
-    def initialize(polling_interval, personal_api_key, project_api_key, host)
+    def initialize(polling_interval, personal_api_key, project_api_key, host, feature_flag_request_timeout_seconds, on_error = nil)
       @polling_interval = polling_interval || 30
       @personal_api_key = personal_api_key
       @project_api_key = project_api_key
@@ -23,6 +23,8 @@ class PostHog
       @group_type_mapping = Concurrent::Hash.new
       @loaded_flags_successfully_once = Concurrent::AtomicBoolean.new
       @feature_flags_by_key = nil
+      @feature_flag_request_timeout_seconds = feature_flag_request_timeout_seconds
+      @on_error = on_error || proc { |status, error| }
 
       @task =
         Concurrent::TimerTask.new(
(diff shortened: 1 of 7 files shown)


lib/posthog/feature_flags.rb
```diff
@@ -161,17 +155,24 @@ def get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {},
...
-          logger.error "Error computing flag locally: #{e}."
+          @on_error.call(-1, "Error computing flag locally: #{e}. #{e.backtrace.join("\n")} ")
           fallback_to_decide = true
         end
       end
       if fallback_to_decide && !only_evaluate_locally
         begin
           flags_and_payloads = get_decide(distinct_id, groups, person_properties, group_properties)
+
+          if !flags_and_payloads.key?(:featureFlags)
```
mariusandra, line 166: 🙈 😆 (not a joke)

```suggestion
          unless flags_and_payloads.key?(:featureFlags)
```

lib/posthog/feature_flags.rb
```diff
@@ -161,17 +155,24 @@ def get_all_flags_and_payloads(distinct_id, groups = {}, person_properties = {},
...
+            raise StandardError.new("Error flags response: #{flags_and_payloads}")
+          end
           flags = stringify_keys(flags_and_payloads[:featureFlags] || {})
           payloads = stringify_keys(flags_and_payloads[:featureFlagPayloads] || {})
         rescue StandardError => e
-          logger.error "Error computing flag remotely: #{e}"
+          @on_error.call(-1, "Error computing flag remotely: #{e}")
+          if raise_on_error
+            raise e
+          end
```
mariusandra, line 175:
```suggestion
          raise if raise_on_error
```

lib/posthog/feature_flags.rb
```diff
@@ -527,9 +539,11 @@ def _request(uri, request_object)
              EOFError,
              Net::HTTPBadResponse,
              Net::HTTPHeaderSyntaxError,
+             Net::ReadTimeout,
+             Net::WriteTimeout,
              Net::ProtocolError => e
         logger.debug("Unable to complete request to #{uri}")
-        throw e
+        raise e
```
mariusandra, line 546:
```suggestion
        raise
```

🙈 🤣

lib/posthog/feature_flags.rb
```diff
@@ -372,6 +373,10 @@ def _compute_flag_locally(flag, distinct_id, groups = {}, person_properties = {}
     def _compute_flag_payload_locally(key, match_value)
       response = nil
 
+      if @feature_flags_by_key.nil?
+        return response
+      end
+
```
mariusandra, line 379:
```suggestion
      return nil if @feature_flags_by_key.nil?

      response = nil
```

lib/posthog/feature_flags.rb
```diff
@@ -46,30 +48,22 @@ def load_feature_flags(force_reload = false)
...
 
-    def get_feature_variants(distinct_id, groups={}, person_properties={}, group_properties={})
-      decide_data = get_decide(distinct_id, groups, person_properties, group_properties)
+    def get_feature_variants(distinct_id, groups={}, person_properties={}, group_properties={}, raise_on_error=false)
+      # TODO: Convert to options hash for easier argument passing
+      decide_data = get_all_flags_and_payloads(distinct_id, groups, person_properties, group_properties, false, raise_on_error)
       if !decide_data.key?(:featureFlags)
-        logger.error "Missing feature flags key: #{decide_data.to_json}"
+        logger.debug "Missing feature flags key: #{decide_data.to_json}"
+        return {}
```
mariusandra, line 56:
```suggestion
        {}
```

mariusandra (approved):
Some ruby thoughts, but 🚢

lib/posthog/feature_flags.rb
neilkakkar: Hmmm, I _thought_ the no return applied only at the end of the block, can I also do this in the middle of the flow?

(this is confusing me, sorry 😅 ). Just for my understanding, what if I had something else on line 59.5 , will this still return early?

lib/posthog/feature_flags.rb
neilkakkar: These are great suggestions, seriously, thank you 😂

lib/posthog/feature_flags.rb
mariusandra: The last executed and thus the last returned line will either be the last line of the `if` block or the last and only line of the `else` block

neilkakkar: oh, that's a better way to think about it, 'the last executed line'. So if at the end of the else block, I added another line that was outside the else block, that would've been returned; but since there is no such line, the last executed one is the last line of the if block. Nice, thanks 👍

PR #47, opened 2025-02-21
feat(feature-flags): support quota limiting for feature flags #85

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide and /local_evaluation APIs if users have gone over their quota limit. Now we need to change the SDKs to handle these new responses.


haacked (approved):
LGTM!

PR #51, opened 2025-04-01
Fix failing test and some other local dev cleanup.

haacked:
This fixes a unit test that was failing due to sending the wrong header in the mocked request.

Other changes make developing on this project a little better. For example, Ruby LSP with Ruby Test Explorer was having issues because when it tries to discover tests and do a dry run, SimpleCov was trying to send results to the server. For whatever reason, the guards to ensure we were not in a dry run were not working. I changed it to check if we're in CI.

Also suppressed output in a test that was spewing out a lot of console output when running tests.

### DIFF
--- a/spec/posthog/feature_flag_spec.rb
+++ b/spec/posthog/feature_flag_spec.rb
@@ -4147,7 +4147,7 @@ class PostHog
             'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
             'Content-Type'=>'application/json',
             'Host'=>'app.posthog.com',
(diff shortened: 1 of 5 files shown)


spec/posthog/feature_flag_spec.rb
```diff
@@ -493,40 +493,6 @@ class PostHog
     end
 
     it 'returns undefined when decide or local_eval times out' do
-      api_feature_flag_res = {
```
haacked, line 496: This variable was defined but not used.

dmarticus (approved):
Phil fixes another repo!

PR #52, opened 2025-04-02
chore(flags): Add request id, version, id, and reason to `$feature_flag_called` event

haacked:
Add support for `/decide?v=4` (see https://github.com/PostHog/posthog/pull/29751). Similar to how it's done in https://github.com/PostHog/posthog-js-lite/pull/427.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

The end result is `$feature_flag_called` events will have additional properties:

| Property Name                |  Descirption
| ------------------------ | -------------
| `$feature_flag_version` | The version of the feature flag. This is visible in the diffs of the feature flag activity log.
| `$feature_flag_reason` | The reason the feature flag matched or didn't match.
| `$feature_flag_id`          | The database id of the feature flag (Just in case a flag was deleted and then recreated with the same key, we can determine the difference.
| $feature_flag_request_id | The ID of the `/decide` request.

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

## Testing

Unit tests and some manual testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.7.3 – 2025-04-03
+## 2.7.3 – 2025-04-07
 
 1. Add more information to `$feature_flag_called` events for `/decide` requests such as flag id, version, reason, and the request id.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.7.3 – 2025-04-03
+
+1. Add more information to `$feature_flag_called` events for `/decide` requests such as flag id, version, reason, and the request id.
+
 ## 2.7.2 – 2025-03-14
 
 1. Fix invocation of shell by ` character
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.7.3 – 2025-04-07
+## 2.8.0 – 2025-04-07
 
 1. Add more information to `$feature_flag_called` events for `/decide` requests such as flag id, version, reason, and the request id.
 
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -77,7 +77,7 @@ class PostHog
         "group_properties": group_properties,
       }
 
-      decide_data = _request_feature_flag_evaluation(request_data)
+      _request_feature_flag_evaluation(request_data)
     end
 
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.7.3 – 2025-04-03
+
+1. Add more information to `$feature_flag_called` events for `/decide` requests such as flag id, version, reason, and the request id.
```
dmarticus, line 3: worth a minor bump vs a patch?  Can't remember what we've done for the other repos.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.7.3 – 2025-04-03
```
dmarticus, line 1:
```suggestion
## 2.7.3 – 2025-04-07
```

CHANGELOG.md
dmarticus, line 3: this is was a minor version bump for other projects, probably worth the same here.

lib/posthog/feature_flags.rb
```diff
@@ -119,9 +139,19 @@ def get_feature_flag(key, distinct_id, groups = {}, person_properties = {}, grou
...
+
       if !flag_was_locally_evaluated && !only_evaluate_locally
         begin
-          flags = get_feature_variants(distinct_id, groups, person_properties, group_properties, true)
+          decide_data = get_all_flags_and_payloads(distinct_id, groups, person_properties, group_properties, false, true)
+          if !decide_data.key?(:featureFlags)
+            logger.debug "Missing feature flags key: #{decide_data.to_json}"
+            flags = {}
+          else
+            flags =stringify_keys(decide_data[:featureFlags] || {})
```
dmarticus, line 151:
```suggestion
            flags = stringify_keys(decide_data[:featureFlags] || {})
```

Do we have a linter?  Feels like one would catch things like this.

lib/posthog/version.rb
```diff
@@ -1,3 +1,3 @@
 class PostHog
-  VERSION = '2.7.2'
+  VERSION = '2.7.3'
```
dmarticus, line 2: same note as above, this is probably worth a `minor` version bump

dmarticus (approved):
code and tests look good; a few comments about versioning and linting.  Nothing blocking.  Nice work.

PR #56, opened 2025-04-28
chore(flags): use new `/flags` endpoint instead of `/decide`

dmarticus:
update Ruby SDK to use the new /flags endpoint instead of the legacy /decide one.


haacked (approved):
:shipit:

Issue #38, opened 2024-02-06
How to send UUID


jclusso:
This [issue](https://github.com/PostHog/posthog/issues/17211) states you must send the `uuid` along with the `event`, `distinct_id`, and `timestamp`. How are you supposed to send the `uuid`?

DataGreed:
posthog doesnt care about their users. This feature exists in python but does not even work, crashing with an api endpoint error `[PostHog] failed to parse request: data did not match any variant of untagged enum RawRequest (400)`

pauldambra:
hey @DataGreed 

> posthog doesnt care about their users.

that cuts deep since we 1000% do... if something isn't working for you the best first step is to request help in the app, the support team can then help route the issue

github comments are, unfortunately, way too noisy for us to get the signal from the noise 😓

pauldambra:
this is resolved in version 2.10.0 which is out at rubygems now

PR #39, opened 2024-02-06
Add `uuid` option to capture

jclusso:
Closes #38 

I couldn't get the test suite to work. Kept getting a ton of errors when running it.

PR #58, opened 2025-05-13
fix: add uuid capture option to capture

pauldambra:
implementation of https://github.com/PostHog/posthog-ruby/pull/39 since it's easier to work locally on my own PR

Closes https://github.com/PostHog/posthog-ruby/issues/38


pauldambra:
@rafaeelaudibert ping on this if  you've time

lib/posthog/field_parser.rb
```diff
@@ -165,6 +171,12 @@ def check_presence!(obj, name)
       def check_is_hash!(obj, name)
         raise ArgumentError, "#{name} must be a Hash" unless obj.is_a? Hash
       end
+
+      def check_is_uuid!(uuid)
```
rafaeelaudibert, line 175: Should we raise or simply log error here? IIRC the JS SDK simply ignores it rather than throwing error

spec/posthog/client_spec.rb
```diff
@@ -419,6 +419,42 @@ class PostHog
...
+            distinct_id: 'distinct_id',
+            event: 'test_event',
+            uuid: '123e4567-e89b-12d3-a456-426614174000'
+          }
+        )
+        last_message = client.dequeue_last_message
+        expect(last_message['uuid']).to eq('123e4567-e89b-12d3-a456-426614174000')
+      end
+
+      it 'generates uuid when not provided' do
```
rafaeelaudibert, line 435: Confusing test name. The UUID is not generated by the gem, but rather by the ingestion pipeline

rafaeelaudibert (approved):
Left some nits, lgtm!

lib/posthog/field_parser.rb
joshsny: Logging and ignoring the passed uuid could be problematic as well leave ingestion to create one - if we log here we should probably fail the capture call and not send it?

joshsny (approved):

lib/posthog/field_parser.rb
pauldambra: wild... you can't provide a UUID to JS Capture I don't think https://github.com/PostHog/posthog-js/blob/chore/csp-violations-playground/src/posthog-core.ts#L902

pauldambra: tension here between two guides i like... https://lawsofux.com/postels-law/ (be tolerant in what you accept and conservative in what you do) and "make the impossible, unrepresentable"

i think maybe logging and not using the not-uuid would be best

capture still works and you get signal

but i can see someone arguing that we should throw and let them discover the problem

🤔

spec/posthog/client_spec.rb
pauldambra:
```suggestion
      it 'does not require a uuid be provided - ingestion will generate when absent' do
```

lib/posthog/field_parser.rb
pauldambra: ok, on balance i think we should log and ignore

you already can't set a UUID so that's not a huge change

we talked about cutting a major release to drop Ruby v2 support, could start to throw then if we want to change to include throwing...

rafaeelaudibert: sounds good to me

PR #61, opened 2025-05-14
chore: Fix all Rubocop errors

rafaeelaudibert:
Let's fix all Rubocop errors. This also uncovered some problems. We're saying in our gemfile we support Ruby 2+ but the code was broken, so fixed it. We should eventually just release a new version limiting us to Ruby 3.2+ (the only supported versions)

Syntax will look much simpler too once we can upgrade (can use `&.` for example)

I've added Ruby 2.0 to the CI to guarantee we dont break this while we dont bump version


pauldambra (approved):
🚀

PR #63, opened 2025-05-20
chore: drop ruby 2 support

### LINKED ISSUE #60: Top-level FeatureFlag class conflicts with existing application classes
Hi PostHog team,

We’ve recently encountered a naming conflict after upgrading to a newer version of the posthog-ruby gem. The gem now defines a top-level FeatureFlag class, which clashes with a class of the same name in our own application.

This can cause unexpected behavior and makes it difficult to use both the gem and our internal codebase without workarounds. As a best practice, Ruby gems typically namespace their classes/modules to avoid such conflicts (e.g., PostHog::FeatureFlag).

Request:
Would you consider moving the FeatureFlag class under a namespace, such as PostHog::FeatureFlag? This would help prevent conflicts for users who have similarly named classes in their projects.

Thank you for your consideration!

pauldambra:
ruby 2 has been end of life for 2 years

https://www.ruby-lang.org/en/downloads/branches/

let's remove support, we've got version 2.10.0 if people really need Ruby 2 support

Closes #60


pauldambra:
interested to see how much of this I can figure out / brute force 🤣

.rubocop.yml
```diff
@@ -1,10 +1,54 @@
 AllCops:
   NewCops: enable
   SuggestExtensions: false
+  TargetRubyVersion: 3.0
```
rafaeelaudibert, line 4: You shouldnt need this because of the `.gemspec` definition

Gemfile
rafaeelaudibert, line 1: Let's update these to their latest? They're all dev/test gems anyway

lib/posthog/client.rb
```diff
@@ -399,12 +400,10 @@ def add_local_person_and_group_properties(distinct_id, groups, person_properties
...
-      if groups
-        groups.each do |group_name, group_key|
-          all_group_properties[group_name] = {
-            :'$group_key' => group_key
-          }.merge((group_properties && group_properties[group_name]) || {})
-        end
+      groups&.each do |group_name, group_key|
+        all_group_properties[group_name] = {
+          '$group_key': group_key
+        }.merge((group_properties && group_properties[group_name]) || {})
```
rafaeelaudibert, line 406: I think this is valid?

```suggestion
        }.merge((group_properties&.[group_name]) || {})
```

pauldambra: ah, no, things were unhappy with that

pauldambra:
ah

https://github.com/PostHog/posthog-ruby/issues/60

we should fix this in this PR too

lib/posthog/client.rb
rafaeelaudibert: It does exist, but it's probably 3.3+ :) ty for trying

rafaeelaudibert (approved):
Pushed 2 small changes, LGTM!

Issue #49, opened 2025-03-03
Disabling sending data in development environment


wjessop:
What's the recommended way of setting the gem up for Rails? I'm looking for:

- Not sending data in development mode (ideally logging instead)
- Making the posthog client available to Ruby code in the app (controllers, models etc.). I see one guide recommends assigning the client to $posthog, but does this work in forking threaded servers such as Puma?

rafaeelaudibert:
There's a new `posthog-rails` gem now, see docs in posthog.com/docs/libraries/ruby-on-rails. You can also disable PostHog in development with a new `test_mode` flag.

PR #64, opened 2025-05-20
feat: add before_send function

pauldambra:
and since I'm here let's add `before_send` to the lib...

would help with e.g. https://github.com/PostHog/posthog-ruby/issues/49

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -350,7 +350,7 @@ class PostHog
     # @param [Object] action The event to be sent to PostHog
     # @return [null, Object, nil] The processed event or nil if the event should not be sent
     def process_before_send(action)
-      return if action.nil? || action.empty?
+      return action if action.nil? || action.empty?
       return action unless @before_send
 
       processed_action = @before_send.call(action)
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -62,7 +62,7 @@ class PostHog
         hash[key] = []
       end
 
-      @before_send = (opts[:before_send] or nil)
+      @before_send = opts[:before_send]
     end
 
     # Synchronously waits until the worker has cleared the queue.
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -26,6 +26,8 @@ class PostHog
     #   Measured in seconds, defaults to 30.
     # @option opts [Integer] :feature_flag_request_timeout_seconds How long to wait for feature flag evaluation.
     #   Measured in seconds, defaults to 3.
+    # @option opts [Proc] :before_send A block that receives the event hash and should return either a modified hash
+    #   to be sent to PostHog or nil to prevent the event from being sent. e.g. `before_send: ->(event) { event }`
     def initialize(opts = {})
       symbolize_keys!(opts)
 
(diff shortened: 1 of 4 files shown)


lib/posthog/client.rb
```diff
@@ -156,7 +160,7 @@ def queued_messages
     end
 
     # TODO: In future version, rename to `feature_flag_enabled?`
-    def is_feature_enabled( # rubocop:disable Naming/PredicateName
+    def is_feature_enabled(# rubocop:disable Naming/PredicateName
```
rafaeelaudibert, line 163: weird?

```suggestion
    def is_feature_enabled( # rubocop:disable Naming/PredicateName
```

lib/posthog/client.rb
```diff
@@ -342,10 +346,31 @@ def shutdown
 
     private
 
+    # before_send should run immediately before the event is sent to the queue.
+    # @param [Object] action The event to be sent to PostHog
+    # @return [null, Object, nil] The processed event or nil if the event should not be sent
+    def process_before_send(action)
+      return if action.nil? || action.empty?
```
rafaeelaudibert, line 353: We're always returning `nil` even in the case where it's simply empty. Maybe should be
```suggestion
      return action if action.nil? || action.empty?
```

lib/posthog/client.rb
```diff
@@ -342,10 +346,31 @@ def shutdown
...
+
+      processed_action = @before_send.call(action)
+
+      if processed_action.nil?
+        logger.warn("Event #{action[:event]} was rejected in beforeSend function")
+      elsif processed_action.empty?
+        logger.warn("Event #{action[:event]} has no properties after beforeSend function, this is likely an error")
+      end
+
+      processed_action
```
rafaeelaudibert, line 364: Maybe we need a begin/rescue here to avoid problems with the `@before_send` call failing. We can then return `nil` in that case

rafaeelaudibert (commented):
a couple comments

lib/posthog/client.rb
```diff
@@ -342,10 +346,31 @@ def shutdown
...
+    end
+
     # private: Enqueues the action.
     #
     # returns Boolean of whether the item was added to the queue.
     def enqueue(action)
+      action = process_before_send(action)
+      return false if action.nil? || action.empty?
```
pauldambra, line 372: technically this is a breaking change
but if someone was relying on the fact that we would enqueue empty objects then 🤷

lib/posthog/client.rb
pauldambra: yeah, i did autoformat in rubymine
silly computers fighting each other i guess

lib/posthog/client.rb
pauldambra: wasn't sure whether to use original or drop on error...

pauldambra: went to check JS SDK and that would explode back at the user 🫠

rafaeelaudibert: It's less bad on JS because `capture` is usually called on useEffects and stuff, but it's much worse in the server where we'll just crash their entire process

rafaeelaudibert (approved):

PR #70, opened 2025-07-22
feat(flags): make the `send_feature_flags` parameter more declarative and ergonomic

dmarticus:
### Problem
  The current `send_feature_flags` boolean parameter provides limited control over feature flag evaluation behavior. Users
  couldn't specify evaluation preferences or custom properties per capture event.

  ### Solution
  Modified `send_feature_flags` to accept either:
  - `bool` (existing behavior, fully backward compatible)
  - `Hash` with granular control options
  - `SendFeatureFlagsOptions` object with granular control:
    ```ruby
    {
      only_evaluate_locally: bool,     # Force local vs remote evaluation
      person_properties: hash,         # Custom properties for this event
      group_properties: hash           # Custom group properties
    }
    ```

  Benefits

  - Declarative control: Explicitly specify local vs remote evaluation per event
  - Custom properties: Supply event-specific properties without affecting global state
  - Backward compatible: Existing boolean usage unchanged
  - Ergonomic: More flexible API that matches user expectations

  Usage Examples

```ruby
  # Traditional usage (unchanged)
  client.capture({
    distinct_id: 'user123',
    event: 'button_clicked',
    send_feature_flags: true
  })

  # New declarative usage with hash
  client.capture({
    distinct_id: 'user123',
    event: 'button_clicked',
    send_feature_flags: {
      only_evaluate_locally: true,
      person_properties: { plan: 'premium' },
      group_properties: { company: { industry: 'tech' } }
    }
  })

  # Using SendFeatureFlagsOptions object
  options = PostHog::SendFeatureFlagsOptions.new(
    only_evaluate_locally: true,
    person_properties: { plan: 'enterprise' }
  )
  client.capture({
    distinct_id: 'user123',
    event: 'button_clicked',
    send_feature_flags: options
  })
```

  Testing

  - Added comprehensive test coverage for all three parameter types
  - Added dedicated tests for the new SendFeatureFlagsOptions class
  - All existing tests continue to pass, confirming backward compatibility

  This brings the Ruby SDK in line with the enhanced Node.js SDK functionality from
  https://github.com/PostHog/posthog-js-lite/pull/566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.1.0
+
+1. feat: make the `send_feature_flags` parameter more declarative and ergonomic
+
 ## 3.0.1 - 2025-05-20
 
 1. fix: was warning on absent UUID when capturing ([#67](https://github.com/PostHog/posthog-ruby/pull/67))


lib/posthog/send_feature_flags_options.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module PostHog
+  # Options for configuring feature flag behavior in capture calls
+  class SendFeatureFlagsOptions
+    attr_reader :only_evaluate_locally, :person_properties, :group_properties
```
haacked, line 6: I see `only_evaluate_locally` defined here, but I don't see the code where we consume this attribute.

lib/posthog/send_feature_flags_options.rb
```diff
@@ -0,0 +1,32 @@
...
+        person_properties: @person_properties,
+        group_properties: @group_properties
+      }
+    end
+
+    def self.from_hash(hash)
+      return nil unless hash.is_a?(Hash)
+
+      new(
+        only_evaluate_locally: hash[:only_evaluate_locally] || hash['only_evaluate_locally'],
```
haacked, line 26: Subtle bug here with how `||` works with "falsy" values. If user passes an explicit `false` value using the symbol:

`hash = { only_evaluate_locally: false }`, 

The expression on the left of `||` results in `false` causing the expression to then evaluate the right side `hash['only_evaluate_locally']`.

Same is true for the next two lines, but not really an issue because a boolean value isn't valid for those. In practice, they'll be an object.

```suggestion
        only_evaluate_locally: hash.key?(:only_evaluate_locally) ?  hash[:only_evaluate_locally] : hash['only_evaluate_locally']
```

This is such a common pattern I wonder if we should have a helper. Something like:

```ruby
# Add to PostHog::Utils module
def get_by_symbol_or_string_key(hash, key)
  symbol_key = key.to_sym
  string_key = key.to_s

  if hash.key?(symbol_key)
    hash[symbol_key]
  else
    hash[string_key]
  end
end
```

haacked (changes requested):
Looks a bit incomplete if I understood the changes correctly.

PR #77, opened 2025-09-22
feat: add exception capture

ablaszkiewicz:
image

Added:
- `capture_exception` in Ruby,
- context lines are supported,
- exception type, exception-like type and string are supported as inputs.

TBD:
- auto capture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 3.3.0
-
-1. feat: add exception capture ([#77](https://github.com/PostHog/posthog-ruby/pull/77))
-
 ## 3.2.0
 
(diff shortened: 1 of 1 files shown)


lib/posthog/exception_capture.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      {
+        '$exception_list' => [exception_info]
+      }
+    end
+
+    def self.build_single_exception(exception)
+      {
+        'type' => exception.class.to_s,
+        'value' => exception.message || '',
```
daibhin, line 32: Ruby is untyped... what happens in the case when a user passes a string rather than an Exception to the `capture_exception` method. I'm guessing this will throw

PR #81, opened 2025-10-28
Add posthog-rails gem for automatic Rails exception tracking

popcorn:
### What is this?

This PR introduces the `posthog-rails` gem which provides automatic exception tracking for Rails applications, inspired by Sentry's gem.

Features:
- Automatic capture of unhandled exceptions via Rails middleware
- Automatic capture of rescued exceptions (configurable)
- Automatic instrumentation of ActiveJob failures
- Integration with Rails 7.0+ error reporter
- Configurable exception exclusion list
- User context capture from controllers

Please read `posthog-ruby/posthog-rails/IMPLEMENTATION.md` to learn more.

### How to test this?

Put both gems in your Rails app's `Gemfile`, and then reference them via path on your local machine:
```ruby
gem "posthog-ruby", path: "/Users/drago/code/posthog-ruby"
gem "posthog-rails", path: "/Users/drago/code/posthog-ruby/posthog-rails"
```

PS
- There's only about ~550 new lines of code, other ~950 lines are readmes, code examples etc.
- I wrote this for myself to test PostHog so I haven't added any specs. If you are happy with this approach let me know and I will write tests.


popcorn:
@daibhin, it seems @hpouillot triggered a few CI actions that are failing.

I'll sort it out tomorrow or during the weekend. In the meantime, let me know if you agree with the general approach.

posthog-rails/examples/posthog.rb
```diff
@@ -0,0 +1,151 @@
...
+#   render 'checkout/old'
+# end
+
+# Capture exceptions manually:
+# begin
+#   dangerous_operation
+# rescue => e
+#   PostHog.capture_exception(e, current_user.id, { context: 'manual' })
+#   raise
+# end
```
rafaeelaudibert: I don't think this makes sense here. This file should be kept as a reasonable default initializer for our customers.

Also, can you add a new `bundle posthog:install` command that automatically copies this file over to `config/initializers/posthog.rb` and document that?

posthog-rails/lib/posthog/rails/active_job.rb
```diff
@@ -0,0 +1,83 @@
...
+      rescue StandardError => e
+        # Capture the exception with job context
+        capture_job_exception(e)
+        raise
+      end
+
+      private
+
+      def capture_job_exception(exception)
+        return unless PostHog.rails_config&.auto_instrument_active_job
```
rafaeelaudibert: Hmmm, I'd much rather this be `PostHog::Rails.config` across the whole PR. Can you update this, please?

posthog-rails/lib/posthog/rails/active_job.rb
```diff
@@ -0,0 +1,83 @@
...
+      def extract_distinct_id_from_job
+        # Try to find a user ID in job arguments
+        arguments.each do |arg|
+          if arg.respond_to?(:id)
+            return arg.id
+          elsif arg.is_a?(Hash) && arg['user_id']
+            return arg['user_id']
+          elsif arg.is_a?(Hash) && arg[:user_id]
+            return arg[:user_id]
+          end
```
rafaeelaudibert: I don't think this makes sense. If someone passes an arbitrary model or entity that returns to `id` that shouldn't be assigned as the `distinct_id` of the USER that's responsible for this action.

We can probably keep the `user_id` hash since it makes some sense. In the future we'll want to introduce some kind of `Posthog.set_context` helper to aid with this.

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -0,0 +1,214 @@
+# frozen_string_literal: true
+
+module PostHog
+  module Rails
+    class Railtie < ::Rails::Railtie
+      # Add PostHog module methods for accessing Rails-specific client
+      initializer 'posthog.set_configs' do |_app|
+        PostHog.class_eval do
```
rafaeelaudibert: As mentioned above, let's configure the config for Rails inside `PostHog::Rails.config` instead - if possible, of course

posthog-rails/IMPLEMENTATION.md
rafaeelaudibert: This is very LLM-y. Do we need this here?

rafaeelaudibert:
Hey @popcorn pinging you just in case you're still interested in getting this merged :)

miharekar:
I'd certainly be interested in getting it merged for my own selfish reasons 😅

johnnagro:
@popcorn no worries, let me see what i can do. thanks for getting a start on this feature.

johnnagro:
@popcorn mind adding me with write permissions to your fork of the repo? so i can push to your branch. seems like the simplest way to continue the PR and maintain all the history

posthog-rails/lib/posthog/rails/active_job.rb
johnnagro: i agree `id` is to aggressive here, remove it. also borrowed from an earlier suggestion and provided a `posthog_distinct_id` DSL for ActiveJob user context. let me know if you think otherwise

posthog-rails/README.md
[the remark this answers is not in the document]
johnnagro: i think we've added a lot more to the docs with all the other updates, let me know if you think we need something specific here

johnnagro:
I was doing some addition testing and saw a couple further changes that needed to be made. I noticed exceptions were being captured twice in PostHog - two `[posthog-ruby] Sending request for 1 items` log lines for each error. I also noticed at times the context serialization was crashing when more complex objects were being serialized or had circular references. I added patches for both.

rafaeelaudibert:
@johnnagro I'll be reviewing this on Monday, thank you!

PR #85, opened 2025-12-10
Add Action workflow for publishing Ruby gem

Piccirello:
This uses trusted publishing.

Screenshot 2025-12-10 at 14 51 23


Piccirello:
There's more work to be done to move the entire versioning and release flow to GitHub Actions, but this is a start.

Issue #88, opened 2026-01-17
Exception capture payload exceeds 32KB limit due to context lines for all stack frames


johnnagro:
I've been testing for #81 and noticed frequent oversized error payloads whenever the error came from within Rails or a gem such as pending migrations.

## Problem

When capturing exceptions, the payload can frequently exceed the 32KB message limit (`Defaults::Message::MAX_BYTES`), causing the message to be silently dropped with only a log warning:

```
[posthog-ruby] a message exceeded the maximum allowed size
```

## Root Cause

In `exception_capture.rb`, the `build_stacktrace` method captures up to 50 stack frames:

```ruby
frames = backtrace.first(50).map do |line|
  parse_backtrace_line(line)
end.compact.reverse
```

For **every frame** where the file exists on disk, `add_context_lines` reads 11 lines of source code (5 pre + 1 current + 5 post):

```ruby
def self.add_context_lines(frame, file_path, lineno, context_size = 5)
  # ...reads actual source code from file
end
```

This is called regardless of whether the frame is `in_app` or from a gem/framework.

**Worst case calculation:**
- 50 frames × 11 lines × ~80 chars/line = ~44KB just for context lines
- Plus frame metadata, exception message, and any additional properties

For Rails applications, the backtrace typically includes many Rails framework frames, so context lines are read from Rails source files as well, bloating the payload.

## Suggested Fixes

1. **Only add context lines for `in_app: true` frames** - Framework/gem code context is less useful and contributes most to size

2. **Reduce default max frames** - 50 is excessive; 20-30 would cover most useful information

3. **Add configurable limits** - Allow users to control:
   - Max frames (`max_exception_frames`)
   - Context line count (`exception_context_lines`)
   - Whether to include context at all (`include_exception_context`)

4. **Add payload size awareness** - Check size before adding to batch, truncate context if needed

## Example Fix for Option 1

```ruby
def self.parse_backtrace_line(line)
  # ... existing parsing ...
  
  # Only add context for in-app frames
  if frame['in_app'] && File.exist?(file)
    add_context_lines(frame, file, lineno)
  end
  
  frame
end
```

## Environment

- Ruby 3.4.5
- Rails 8.1.1
- posthog-ruby (local/dev version)

Issue #87, opened 2026-01-17
Empty batch sent after oversized message rejection causes retry loop with JSON parse error


johnnagro:
While working on #81 I noticed some logging and investigated.

When a message exceeds the maximum allowed size (32KB), it is correctly rejected by `MessageBatch`, but two subsequent issues cause a noisy retry loop.

## Issue 1: Empty batch is sent to API

In `send_worker.rb`, when all messages in a batch are rejected due to size limits, the worker still calls `@transport.send` with an empty batch:

```ruby
# send_worker.rb:46
res = @transport.send @api_key, @batch  # @batch may be empty here
```

The log shows:

```
[posthog-ruby] a message exceeded the maximum allowed size
[posthog-ruby] Sending request for 0 items
```

## Issue 2: JSON parsing fails on non-JSON response

When the server receives an invalid/empty request, it may return a non-JSON error response. The transport code assumes JSON:

```ruby
# transport.rb:53
error = JSON.parse(body)['error']
```

This throws a `JSON::ParserError`, which triggers the retry logic, resulting in:

```
[posthog-ruby] Retrying request, 10 retries left
[posthog-ruby] Retrying request, 9 retries left
...
[posthog-ruby] unexpected character: 'request' at line 1 column 1
```

## Suggested fixes

1. Skip sending if batch is empty in `send_worker.rb`
2. Wrap JSON parsing in a rescue block in `transport.rb`

## Environment

- Ruby 3.4.5
- Rails 8.1.1
- posthog-ruby (local/dev version)

rafaeelaudibert:
Closed by #98

PR #98, opened 2026-02-06
fix: Fix Railtie middleware insertion crashing on Rails initialization

### LINKED ISSUE #97: NoMethodError: undefined method 'insert_middleware_after' in Railtie initializer
The posthog.insert_middlewares initializer in PostHog::Rails::Railtie calls insert_middleware_after without a receiver. Inside an initializer block, self is the Railtie instance, but insert_middleware_after is defined as a class method (def self.insert_middleware_after), causing a NoMethodError at boot time.
The same issue affects register_error_subscriber and rails_version_above_7? in the config.after_initialize block.

Steps to reproduce:

1. Add posthog-rails (3.5.1) to a Rails 8.1 app

1. Run rails generate posthog:install (or any command that boots the app)

```
~> rails generate posthog:install
/Users/johnnagro/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/posthog-rails-3.5.1/lib/posthog/rails/railtie.rb:79:in 'block in <class:Railtie>': undefined method 'insert_middleware_after' for an instance of PostHog::Rails::Railtie (NoMethodError)

        insert_middleware_after(
        ^^^^^^^^^^^^^^^^^^^^^^^
```

johnnagro:
fixes #97, #87, #88

## Fixes initializer middleware errors #97

The initializer block called insert_middleware_after as a class method, but Rails runs initializer blocks via instance_exec on the Railtie instance. Also removed the include? check on MiddlewareStackProxy which doesn't support query methods during initialization.

I created a fresh rails project, added `posthog-ruby` and the local copy of this `posthog-rails` branch. Then verified that `bin/rails generate posthog:install` runs cleanly again.

## Fixes Exception capture payload & Empty Batch Errors #87

Also fixes when a message exceeds the 32KB size limit, it is silently dropped from the batch. Previously the worker would still send the empty batch to the API, which returned a non-JSON response, causing a JSON::ParserError
that triggered 10 pointless retries.

- Skip transport request when the batch is empty after consuming the queue
- Rescue JSON::ParserError in Transport#send and fall back to the raw
  response body instead of crashing

## Only include source context lines for in-app exception frames #88

Exception payloads frequently exceeded the 32KB message limit because
context lines (11 lines of source code) were read for every stack frame,
including gem and framework frames. A typical Rails backtrace has ~50
frames, most from Rails internals, easily blowing past the limit and
causing the event to be silently dropped.

Now context lines are only added for in_app frames. Gem/framework frames
still include file, line number, and function name.

## Fixes Use $current_url property so exception URL appears in PostHog UI

The CaptureExceptions middleware was setting $request_url, which is not
recognized by PostHog for the "URL / Screen" column. Renamed to
$current_url to match the standard property used across PostHog SDKs.

Screen Shot 2026-02-06 at 3 45 49 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## 3.5.3 - 2026-02-08
 
-1. fix: Fix Railtie middleware insertion crashing on Rails initialization — changed `insert_middleware_after` from a class method to an instance method (matching how Rails executes initializer blocks via `instance_exec`), and removed the unsupported `include?` query on `MiddlewareStackProxy`
-2. fix: Prevent sending empty batches and handle non-JSON response bodies gracefully in transport layer
+1. fix: Fix Railtie middleware insertion crashing on Rails initialization — changed `insert_middleware_after` from a class method to an instance method (matching how Rails executes initializer blocks via `instance_exec`), and removed the unsupported `include?` query on `MiddlewareStackProxy` ([#97](https://github.com/PostHog/posthog-ruby/issues/97))
+2. fix: Prevent sending empty batches and handle non-JSON response bodies gracefully in transport layer ([#87](https://github.com/PostHog/posthog-ruby/issues/87))
 3. fix: Use `$current_url` property (instead of `$request_url`) so exception URLs appear correctly in the PostHog UI
-4. fix: Only include source context lines for in-app exception frames, avoiding unnecessary reads of gem source files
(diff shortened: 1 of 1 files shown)


johnnagro:
@rafaeelaudibert i'm happy to help. I've made the changes you've requested. At your convenience please take another look and let me know if you'd like me to make any additional tweaks.

PR #137, opened 2026-04-27
feat: add evaluate_flags() API for single-call flag evaluation

dmarticus:
## Summary

- Adds `Client#evaluate_flags(distinct_id, …)` returning a `FeatureFlagEvaluations` snapshot — one `/flags` round-trip powers both branching and event enrichment.
- Snapshot exposes `is_enabled` / `get_flag` / `get_flag_payload` for branching, plus `only_accessed` / `only([keys])` to narrow what gets attached to a captured event.
- Adds `flags:` option on `capture` and `capture_exception` — when present, attaches `$feature/<key>` and `$active_feature_flags` from the snapshot without any extra `/flags` call.
- `flag_keys:` on `evaluate_flags` scopes the underlying `/flags` request itself (sent as `flag_keys_to_evaluate`).
- Deprecates `is_feature_enabled`, `get_feature_flag`, `get_feature_flag_result`, `get_feature_flag_payload`, and `capture(send_feature_flags:)` — they keep working but now emit a `DeprecationWarning` pointing at `evaluate_flags()`. Removal is planned for the next major.
- Adds a `feature_flags_log_warnings:` client option (default `true`) to silence the snapshot's filter-helper warnings.

## References

RFC: https://github.com/PostHog/requests-for-comments-internal/pull/1020 · mirrors [posthog-python#539](https://github.com/PostHog/posthog-python/pull/539) and [posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476).

## Design decisions

- `is_enabled` returns `false` for unknown flags, `get_flag` returns `nil` — matches the legacy single-flag methods so existing branching code is structurally interchangeable.
- `get_flag_payload` deliberately does **not** record access or fire `$feature_flag_called` — payload-only reads shouldn't count as an exposure.
- `only_accessed()` returns an empty snapshot when nothing has been accessed (it honors its name) — pre-access flags first if you want a populated result.
- Filtered snapshot clones get a copy of the parent's `accessed` set, so calls on a clone don't back-propagate exposures into the parent.
- A small `Host` struct (two lambdas: `capture_flag_called_event_if_needed`, `log_warning`) is passed to the snapshot instead of a back-reference to the full `Client` — keeps the snapshot decoupled and testable.
- Locally-evaluated flags are tagged with `locally_evaluated: true`, reason `"Evaluated locally"`, and `$feature_flag_definitions_loaded_at` on emitted events, matching the existing single-flag local path.
- Response-level errors (`errorsWhileComputingFlags`, `quotaLimited`) and per-flag errors (`flag_missing`) are combined into a comma-joined `$feature_flag_error` so each event keeps the granular error code(s).
- Passing both `flags:` and `send_feature_flags:` to `capture` uses the snapshot and warns — the snapshot guarantees the event carries the same values your code branched on.
- Deprecated methods bypass each other internally (`is_feature_enabled` and `get_feature_flag` call a private `_get_feature_flag_result` directly) so a single user-level call emits exactly one warning, not a cascade.

## Phase 2 follow-ups

- Update docs at https://posthog.com/docs/libraries/ruby and the feature-flags pages so the wizard / new-app instrumentation points at `evaluate_flags()`.
- Consider stashing the active `flags` snapshot on a request-scoped context so auto-captured exceptions inherit the same flag context as manual captures.

All tests + lint pass on Ruby 3.2 / 3.3 / 3.4 (380 examples).

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*


lib/posthog/feature_flag_evaluations.rb
```diff
@@ -0,0 +1,173 @@
...
+      @errors_while_computing = errors_while_computing
+      @quota_limited = quota_limited
+      @accessed = Set.new(accessed || [])
+    end
+
+    def keys
+      @flags.keys
+    end
+
+    def is_enabled(key) # rubocop:disable Naming/PredicateName
```
dustinbyrne: `enabled?(key)` feels more Ruby idiomatic here

lib/posthog/feature_flag_evaluations.rb
```diff
@@ -0,0 +1,173 @@
...
+      key = key.to_s
+      flag = @flags[key]
+      response = flag&.enabled ? true : false
+      _record_access(key, flag, response)
```
dustinbyrne: For variant flags, this will emit the `$feature_flag_called` event with `$feature/variant-flag: true | false` instead of the variant. This means `is_enabled` and `get_flag` will both emit `$feature_flag_called` due to the different response.

dustinbyrne: similarly, response will be `false` instead of `nil` if the flag is missing

lib/posthog/client.rb
```diff
@@ -135,6 +136,7 @@ def initialize(opts = {})
       end
 
       @before_send = opts[:before_send]
+      @feature_flags_log_warnings = opts.key?(:feature_flags_log_warnings) ? opts[:feature_flags_log_warnings] : true
```
dustinbyrne: Do we need this, or should we gate via log level?
```ruby                                                                                                                                                                                                             
config.log_level = :debug
```

similar to https://github.com/PostHog/posthog-android/pull/498#discussion_r3169888176

lib/posthog/client.rb
```diff
@@ -183,8 +189,31 @@ def clear
...
+
       send_feature_flags_param = attrs[:send_feature_flags]
       if send_feature_flags_param
+        Kernel.warn(
+          '`send_feature_flags` is deprecated and will be removed in a future major version. ' \
+          'Pass a `flags` snapshot from `client.evaluate_flags(...)` instead — it avoids a ' \
+          'second `/flags` request per capture and guarantees the event carries the exact ' \
+          'flag values your code branched on.',
+          category: :deprecated, uplevel: 1
+        )
```
dustinbyrne: Hmm, we don't have a great way of presenting deprecations in this SDK. 

> [Deprecation warnings are no longer shown by default (since Ruby 2.7.2). Turn them on with -W:deprecated (or with -w to show other warnings too)](https://docs.ruby-lang.org/en/3.4/NEWS/NEWS-3_0_0_md.html)

Realistically I don't know if anyone is going to see this. Maybe we drop the `:deprecated` category and only log once per unique method? [ActiveSupport has it's own deprecation API](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html), so I don't think we'd be stepping outside of what's acceptable.

lib/posthog/client.rb
```diff
@@ -414,18 +462,137 @@ def get_feature_flag_result(
...
+        locally_evaluated_keys << key.to_s
+      end
+
+      request_id = nil
+      evaluated_at = nil
+      errors_while_computing = false
+      quota_limited = false
+
+      unless only_evaluate_locally
```
dustinbyrne: We might have matched all the flags locally already, but unless `only_evaluate_locally` is true we'll still hit /flags

lib/posthog/client.rb
```diff
@@ -414,18 +462,137 @@ def get_feature_flag_result(
...
+            key_str = key.to_s
+            next if locally_evaluated_keys.include?(key_str)
+
+            metadata = ff.metadata
+            reason = ff.reason
+            records[key_str] = FeatureFlagEvaluations::EvaluatedFlagRecord.new(
+              key: key_str,
+              enabled: ff.enabled ? true : false,
+              variant: ff.variant,
+              payload: ff.payload,
```
dustinbyrne: We should normalize the payload to a deserialized type

lib/posthog/client.rb
```diff
@@ -414,18 +462,137 @@ def get_feature_flag_result(
...
+      request_id = nil
+      evaluated_at = nil
+      errors_while_computing = false
+      quota_limited = false
+
+      unless only_evaluate_locally
+        begin
+          flags_response = @feature_flags_poller.get_flags(
+            distinct_id, groups, person_properties, group_properties, flag_keys
+          )
```
dustinbyrne, line 533: We accept `disable_geoip` as a param but don't do anything with it. `get_flags` doesn't look to support it.

lib/posthog/client.rb
```diff
@@ -414,18 +462,137 @@ def get_feature_flag_result(
...
+      )
+
+      records = {}
+      locally_evaluated_keys = Set.new
+
+      @feature_flags_poller.load_feature_flags
+      poller_flags_by_key = @feature_flags_poller.feature_flags_by_key || {}
+
+      poller_flags_by_key.each do |key, definition|
+        next if flag_keys && !flag_keys.map(&:to_s).include?(key.to_s)
```
dustinbyrne: Prefer using a set here

lib/posthog/client.rb
```diff
@@ -183,8 +189,31 @@ def clear
...
+      # `send_feature_flags`. The snapshot guarantees the event carries the same
+      # values the developer branched on with no additional network call.
+      if attrs[:flags]
+        if attrs[:send_feature_flags]
+          logger.warn(
+            '[FEATURE FLAGS] Both `flags` and `send_feature_flags` were passed to ' \
+            'capture(); using `flags` and ignoring `send_feature_flags`.'
+          )
+        end
+        snapshot_props = attrs[:flags]._get_event_properties
```
dustinbyrne: We should guard against the `NoMethodError` here

lib/posthog/client.rb
```diff
@@ -386,25 +418,41 @@ def get_feature_flag_result(
...
+        key, distinct_id,
+        groups: groups, person_properties: person_properties, group_properties: group_properties,
+        only_evaluate_locally: only_evaluate_locally, send_feature_flag_events: send_feature_flag_events
+      )
+    end
+
+    # Internal: implementation of {#get_feature_flag_result}, called by both the
+    # public method and the legacy `is_feature_enabled` / `get_feature_flag`
+    # paths. Bypassing the public wrapper avoids cascading deprecation warnings.
+    def _get_feature_flag_result(
```
dustinbyrne: This is in the public section, can it be moved to private?

dustinbyrne (approved):

PR #147, opened 2026-05-12
fix: accept symbol feature flag keys

dustinbyrne:
Feature flag APIs should accept Ruby symbols for flag keys as well as strings. Previously, symbol keys could miss local and remote flag lookups because SDK internals compared them directly against string keys.

This normalizes flag keys to strings across the flag APIs, including legacy single-flag methods, remote config payloads, and `evaluate_flags(flag_keys:)`.

Review loop: checked the SDK diff against compatibility, API surface, entry-point tracing, performance, coverage, and error-handling criteria; no follow-up fixes were needed after the review.

- `bundle exec rspec`
- `bundle exec rubocop`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -364,15 +364,15 @@ module PostHog
       !!response
     end
 
-    # @param [String] flag_key The unique flag key of the feature flag
+    # @param [String, Symbol] flag_key The unique flag key of the feature flag
     # @return [String] The decrypted value of the feature flag payload
     def get_remote_config_payload(flag_key)
-      @feature_flags_poller.get_remote_config_payload(flag_key)
+      @feature_flags_poller.get_remote_config_payload(flag_key.to_s)
     end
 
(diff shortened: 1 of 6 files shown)


lib/posthog/client.rb
```diff
@@ -364,15 +364,15 @@ def is_feature_enabled( # rubocop:disable Naming/PredicateName
       !!response
     end
 
-    # @param [String] flag_key The unique flag key of the feature flag
+    # @param [String, Symbol] flag_key The unique flag key of the feature flag
```
marandaneto, line 367: would that be a breaking change?

dustinbyrne: No, it's additive. This is like a `String | Symbol` in TypeScript, so you can now optionally call using a symbol like:
```rb
get_feature_flag(:beta_flag)
```

marandaneto (approved):

PR #163, opened 2026-06-02
feat: emit $is_server property on captured events

turnipdabeets:
Adds `$is_server: true` to every event captured by this SDK so PostHog ingestion can identify server-side events.

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -73,6 +73,9 @@ module PostHog
     #   Intended only for local development or custom deployments.
     # @option opts [Object] :flag_definition_cache_provider An object implementing the {FlagDefinitionCacheProvider}
     #   interface for distributed flag definition caching.
+    # @option opts [Boolean] :is_server +true+ to stamp captured events with `$is_server => true` so PostHog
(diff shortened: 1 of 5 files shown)


lib/posthog/client.rb
```diff
@@ -334,6 +340,7 @@ def group_identify(attrs)
       return false if @disabled
 
       symbolize_keys! attrs
+      attrs[:is_server] = @is_server
       enqueue(FieldParser.parse_for_group_identify(attrs))
```
dustinbyrne, line 344: `FieldParser.parse_for_group_identify` writes attrs to an inner `$group_set` property instead of a top level property, so this won't work

dustinbyrne: worth testing the shape of these enqueued events to catch this

turnipdabeets: updated

lib/posthog/field_parser.rb
```diff
@@ -89,11 +89,11 @@ def parse_for_group_identify(fields)
         common.merge(
           {
             event: '$groupidentify',
-            properties: {
+            properties: (common[:properties] || {}).merge(
```
dustinbyrne, line 92: nice bug fix

dustinbyrne (approved):

PR #166, opened 2026-06-04
feat(feature-flags): support early_exit in local evaluation

gustavohstrassburger:

PostHog added an "early exit" option to feature flags in the server-side (Rust) evaluation engine, and it has already been ported to posthog-node and posthog-python. This PR ports the same behavior to the **Ruby SDK's local evaluation**.

When a flag's `filters.early_exit` is `true`, local condition evaluation must STOP and return a definitive disabled result as soon as a condition group's property filters match (or the group has no property filters) but the rollout percentage EXCLUDES the user — instead of falling through to evaluate later condition groups.

To mirror the Rust engine exactly, the single-group matcher now returns a tri-state outcome instead of a boolean:

- **`:match`** — property filters matched AND rollout included the user (existing behavior).
- **`:no_match`** — a property filter did NOT match -> always falls through to the next group, even with `early_exit` enabled.
- **`:out_of_rollout_bound`** — property filters matched (or there were none) but the rollout percentage excluded the user. With `early_exit` enabled, the loop returns a definitive disabled result immediately; otherwise existing fall-through behavior is preserved.

`early_exit` is read from the flag's `filters` and defaults to `false`/absent, so behavior is unchanged for existing flags.

- Added unit tests against `match_feature_flag_properties` covering: (a) `early_exit` enabled returns `false` without evaluating a later group that would match; (b) `early_exit` unset falls through to the later matching group (returns `true`); (c) `early_exit` explicitly `false` falls through (returns `true`); (d) a group failing on a property filter (not rollout) does NOT early-exit even when enabled and falls through to a later matching group. The first group uses matching properties with `rollout_percentage: 0` and the second uses matching properties with `rollout_percentage: 100`.
- Existing `condition_match` tests still pass via a thin boolean wrapper over the new tri-state method.
- Ran the full suite: `bundle exec rspec` -> 446 examples, 0 failures.
- Ran the linter/formatter: `bundle exec rubocop` -> no offenses on changed files.

## :pencil: Checklist

### If releasing new changes

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

### DIFF
--- a/.changeset/feature-flag-early-exit.md
+++ b/.changeset/feature-flag-early-exit.md
@@ -1,0 +1,5 @@
+---
+"posthog-ruby": minor
+---
+
+Support the `early_exit` option in local feature flag evaluation. When a flag's `filters.early_exit` is `true`, evaluation stops and returns a definitive disabled result as soon as a condition group's property filters match (or it has none) but the rollout percentage excludes the user, instead of falling through to later condition groups. This mirrors the server-side evaluation engine (and posthog-node / posthog-python). Property-mismatch groups still fall through as before, and behavior is unchanged when `early_exit` is unset or `false`.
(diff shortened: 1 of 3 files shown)


dustinbyrne (approved):
greptile loves parameterization but this looks good

PR #167, opened 2026-06-04
feat(rails): forward Rails.logger to PostHog Logs via OpenTelemetry

johnnagro:
Adds an opt-in integration that forwards `Rails.logger` output to PostHog Logs
(https://posthog.com/docs/logs) over OpenTelemetry/OTLP, automatically correlated
with the request's distinct_id and session_id captured by the existing
request-context middleware.
The OpenTelemetry gems are optional/soft dependencies loaded lazily, so the
feature no-ops with a single warning when absent (and on Ruby < 3.3 where the
logs SDK is unsupported). The logs token and host are derived from the configured
client (`config.api_key` / `config.host`) with ENV fallbacks, so logs always
target the same project as analytics. Disabled by default; enabled via
`PostHog::Rails.config.logs_enabled`.
- Unit specs for the appender, setup (gems present/absent + token/host resolution),
  and railtie broadcast wiring (full suite green on Ruby 3.2.2, rubocop clean).
- End-to-end in an ephemeral `rails new` app against a real PostHog project:
  `Rails.logger` output appears in PostHog Logs. <add screenshots>
## :pencil: Checklist

Screen Shot 2026-06-04 at 7 24 04 PM


posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,199 @@
...
+            provider.add_log_record_processor(processor)
+            provider
+          end
+
+          def resource_attributes(config)
+            attrs = {
+              'service.name' => service_name,
+              'service.version' => PostHog::VERSION,
+              'deployment.environment' => ::Rails.env.to_s
+            }
```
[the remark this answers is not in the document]
johnnagro: Instead, I will just drop the service.version line and keep service.name + deployment.environment. That removes the incorrect attribute without adding a new questionable one, and the gem version still travels with each record via the instrumentation scope. Change incoming.

turnipdabeets:
Cc @PostHog/logs

spec/posthog/rails/logs/appender_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(record[:severity_text]).to eq('INFO')
+    end
+
+    it 'maps error severity' do
+      appender.error('boom')
+
+      record = otel_logger.emitted.first
+      expect(record[:severity_number]).to eq(17)
+      expect(record[:severity_text]).to eq('ERROR')
+    end
```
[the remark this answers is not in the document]
johnnagro: fair suggestion, I'll replace the single maps error severity test with a data-driven test covering all six mappings. Change incoming.

lib/posthog/client.rb
```diff
@@ -51,6 +51,13 @@ def _decrement_instance_count(api_key)
       end
     end
 
+    # @return [String, nil] The project API key this client was initialized with
+    #   (after whitespace trimming). Nil when the client is disabled.
```
turnipdabeets: is there an alternative here that doesn't add to the public API? what about env/config instead?

turnipdabeets: I think this is also stored raw, never trimmed, can you confirm?

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,105 @@
...
+      # `Rails.logger` so that ordinary `Rails.logger.info(...)` calls flow to
+      # PostHog Logs in addition to the normal output. Each record is stamped
+      # with the request-scoped PostHog identity captured by
+      # {PostHog::Rails::RequestContext}.
+      #
+      # @api private
+      class Appender < ::Logger
+        SELF_LOG_PREFIX = '[posthog-ruby]'
+        SELF_LOG_PROGNAME = 'PostHog'
+        REQUEST_ATTRIBUTE_KEYS = %w[$current_url $request_method $request_path].freeze
```
turnipdabeets: I think we can also add `feature_flags`

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,105 @@
...
+        rescue StandardError
+          # Never let log forwarding break the calling code path.
+          true
+        end
+
+        private
+
+        def emit(severity, message, progname)
+          severity_number, severity_text = Severity.for(severity)
+          @otel_logger.on_emit(
```
turnipdabeets: No explicit thread-safety here

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,203 @@
...
+            )
+            false
+          end
+
+          def build_provider(config, token)
+            resource = OpenTelemetry::SDK::Resources::Resource.create(resource_attributes(config))
+            provider = OpenTelemetry::SDK::Logs::LoggerProvider.new(resource: resource)
+            exporter = OpenTelemetry::Exporter::OTLP::Logs::LogsExporter.new(
+              endpoint: logs_endpoint(resolve_host),
+              headers: { 'Authorization' => "Bearer #{token}" }
```
turnipdabeets, line 174: @PostHog/logs js + other sdks use `?token=` vs `Bearer auth` - is there a preference?

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,105 @@
...
+            attributes[key] = value if value
+          end
+
+          attributes
+        end
+
+        def self_log?(message, progname)
+          return true if progname.to_s == SELF_LOG_PROGNAME
+
+          message.is_a?(String) && message.include?(SELF_LOG_PREFIX)
```
turnipdabeets: I wonder if this should be `start_with?` instead of `include?`

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
+
+require 'logger'
+require 'posthog/logging'
+require 'posthog/rails/logs/appender'
+
```
turnipdabeets, line 8: @PostHog/logs other Sdks use JSON vs protobuf - is there a preference?

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,203 @@
...
+          end
+
+          def build_provider(config, token)
+            resource = OpenTelemetry::SDK::Resources::Resource.create(resource_attributes(config))
+            provider = OpenTelemetry::SDK::Logs::LoggerProvider.new(resource: resource)
+            exporter = OpenTelemetry::Exporter::OTLP::Logs::LogsExporter.new(
+              endpoint: logs_endpoint(resolve_host),
+              headers: { 'Authorization' => "Bearer #{token}" }
+            )
+            processor = OpenTelemetry::SDK::Logs::Export::BatchLogRecordProcessor.new(exporter)
```
turnipdabeets, line 176: I think we want a rate cap otherwise we can burn through the quota

posthog-rails/examples/posthog.rb
```diff
@@ -43,6 +43,32 @@
   #   # 'MyCustom404Error',
   #   # 'MyCustomValidationError'
   # ]
+
+  # --------------------------------------------------------------------------
+  # POSTHOG LOGS (OpenTelemetry) - opt-in
+  # --------------------------------------------------------------------------
+  # Forward Rails.logger output to PostHog Logs over OTLP, automatically
```
turnipdabeets, line 50: I think it would be good to have a `beforeSend` to scrub PII.

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -118,12 +118,20 @@ def ensure_initialized!
         register_error_subscriber if rails_version_above_7?
       end
 
+      # Opt-in: forward logs to PostHog Logs over OTLP
+      config.after_initialize do
+        install_posthog_logs if PostHog::Rails.config&.logs_enabled
+      end
```
turnipdabeets, line 128: build_provider runs in after_initialize (pre-fork in preloaded servers), and install! is idempotent so workers can't rebuild it. Does BatchLogRecordProcessor self-restart its thread via Process._fork? If not, logs are captured but never flushed from forked workers — we'd need to build/reset the provider in an after-fork hook instead. Was the E2E run single-process or cluster mode?

posthog-rails/lib/posthog/rails/logs/setup.rb
DanielVisca: for `?token=` vs `Authorization: Bearer` Use `Authorization: Bearer`, since our ingestion service checks it first and all docs treat it as primary (`?token=` is just a fallback).

posthog-rails/lib/posthog/rails/logs/setup.rb
DanielVisca: Preference is protobuf, since the endpoint decodes it first and it's the OTel default (JSON is only required in the Next.js/edge guide).

posthog-rails/lib/posthog/rails/logs/setup.rb
DanielVisca: Reading the spec, since logs CAN accept either, we want to avoid exceptions, stick with otel or rails standards here

lib/posthog/client.rb
johnnagro: Confirmed it's trimmed. Initialize runs `opts[:api_key] = normalize_string_option(opts[:api_key])` (which calls `.strip`) before `@api_key = opts[:api_key]`, so the reader returns the normalized value. However, the doc wording is imprecise when it says "Nil when disabled" since a blank key disables the client but stores "". I'll fix the comment (or this becomes moot per your other comment).

johnnagro: Yes, good point. I've reworked it so the core client is untouched. `PostHog.init` (defined in posthog-rails) already sees the resolved options before constructing the client, so the logs setup now captures api_key/host there, with the existing ENV fallbacks for anyone constructing a client directly. The attr_readers on `PostHog::Client` will be reverted.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Really cool idea. One question first: the appender forwards attributes from the request context, and right now nothing populates feature flags into that context automatically. Flag evaluations stay in the returned snapshot and only attach to capture events when passed explicitly. For this to work i think we'd also need to wire flag evaluation into the request context. 

Is that the desired state here, or a later addition?

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: i believe this is correct. `emit` has no shared mutable state. Synchronization happens downstream in OTel's BatchLogRecordProcessor. I added a comment to clarify. Let me know if you disagree though - we need these things to be correct.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: thanks. i think we'll leave it as is.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: good call.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Agreed this was worth doing. I've added an opt-out rate cap. `config.logs_max_records_per_minute` (default 6,000, nil disables) is enforced in the appender via a fixed one-minute window (monotonic clock, mutex-guarded counter). The check runs after level filtering and self-log suppression, so only records that would actually ship count against the cap. When it trips, a single WARN record is emitted noting that further records are being dropped for the remainder of the window, so truncation is never silent.

This mirrors what the other PostHog SDKs with built-in logs support already do: posthog-react-native ships a tumbling-window rateCap (default 500/10s) and posthog-js a maxLogsPerInterval (default 1,000 per 3s flush window), both with a single warning on drop. The 6,000/min default sits between the two, happy to adjust the number if you have a preference.

posthog-rails/examples/posthog.rb
johnnagro: Cool idea. The RN SDK's logs config has exactly this (logs.beforeSend), and the core Ruby client already has before_send for events, so I added `config.logs_before_send` with matching semantics: the proc receives a record hash (`:body`, `:severity`, `:attributes`, `:timestamp`) and returns a modified hash to send or nil to drop. One deliberate difference from the events hook: if the callback raises, the record is dropped rather than sent unmodified. I thought this was the right way if we're thinking about  a PII scrubber, failing closed seemed like the right default. It also runs after the rate-cap check, so a log flood doesn't pay scrubbing costs for records that would be dropped anyway.

posthog-rails/lib/posthog/rails/railtie.rb
johnnagro: Yeah, great question - this is handled inside the OTel SDK. `BatchLogRecordProcessor` does its own fork detection ([source](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/logs_sdk/lib/opentelemetry/sdk/logs/export/batch_log_record_processor.rb)). There's `OTEL_RUBY_BLRP_START_THREAD_ON_BOOT=false` for preload setups that want to skip the master's thread entirely.

The E2E run was single-process (dev-mode Puma), so it didn't exercise this path directly - happy to add a cluster-mode smoke test if you'd like.

johnnagro:
Thanks for taking the time to look this over and leave such thoughtful comments, I appreciate it @turnipdabeets @DanielVisca

I made some changes based on your feedback. There is one open question about feature flags.

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+              "PostHog Logs rate cap reached (#{@rate_limiter.limit} records/minute); " \
+              'dropping further records for the remainder of this window',
+              nil
+            )
+            return true
+          end
+
+          emit(severity, message, progname)
+          true
+        rescue StandardError
```
turnipdabeets: Two things hiding in this rescue:

1. It can't catch `SystemStackError`. If a `logs_before_send` callback calls `Rails.logger` the broadcast re-enters add, recursion blows the stack before the rate cap can stop it, and the SystemStackError
  escapes this rescue and 500s every request that logs. A thread-local re-entrancy guard might fix it.

2. If the emit path fails persistently, this rescue silently drops 100% of logs forever with no warning anywhere. The raise path in apply_before_send warns once; same here would make it debuggable.

```ruby
def add(severity, message = nil, progname = nil)
    return true if Thread.current[:posthog_logs_emitting]

    begin
      Thread.current[:posthog_logs_emitting] = true
      # ... existing body unchanged ...
      true
    rescue StandardError => e
      warn_emit_error(e)  # warn-once, same shape as warn_before_send_error
      true
    ensure
      Thread.current[:posthog_logs_emitting] = nil
    end
  end
```

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+      # limiter's counter. Emitting touches no shared mutable state
+      # (`@otel_logger` is assigned once, attributes are built per call, and
+      # `Internal::Context.current` is thread/fiber-local), and the OTel
+      # BatchLogRecordProcessor synchronizes its buffer internally — the same
+      # split as stdlib `Logger`, which locks in `LogDevice`, not
+      # `Logger#add`. A mutex around emit would serialize all app logging
+      # needlessly.
+      #
+      # @api private
+      class Appender < ::Logger
```
turnipdabeets, line 31: I think storing the threshold in Logger#level has side effects on Rails 7.1+ because BroadcastLogger shares the level across sinks:

  - logs_level = :debug makes Rails.logger.debug? true app-wide, so ActiveRecord for example starts logging every SQL query — and those lines ship only to PostHog (the file logger is still at :info). 
  - And the reverse: Rails.logger.level = :error overwrites the configured logs_level.

  Cheap fix: keep the threshold in a private ivar checked in #add, leave Logger#level alone.

  Two smaller gaps, we should probably document: Rails.logger.silence doesn't silence the appender (muted logs still ship), and tags are lost — config.log_tags request IDs never reach records, and Rails.logger.tagged('X') (non-block form) returns a logger that skips the appender entirely.

wdyt?

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,225 @@
...
+          rescue StandardError
+            'rails'
+          end
+
+          def logs_endpoint(host)
+            base = (host || 'https://us.i.posthog.com').to_s.sub(%r{/+\z}, '')
+            "#{base}/i/v1/logs"
+          end
+
+          def resolve_level(level)
```
turnipdabeets, line 207: A typo like logs_level = :warning (Python/syslog spelling) hits the NameError rescue and silently inherits the Rails logger level instead — the user thinks they're filtering to warnings and quietly forwards :info/:debug volume. A warn_once naming the bad value (and the valid symbols) would match how every other misconfig in this file speaks up.

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,225 @@
...
+            )
+          rescue StandardError => e
+            warn_once("Failed to initialize PostHog Logs: #{e.message}")
+            nil
+          end
+
+          # Flush any buffered log records.
+          #
+          # @return [void]
+          def force_flush
```
turnipdabeets: Does `force_flush` have any callers, can it be removed?

posthog-rails/lib/posthog/rails/configuration.rb
```diff
@@ -33,6 +33,30 @@ class Configuration
...
+      attr_accessor :logs_max_records_per_minute
+
+      # @return [Proc, nil] Callback invoked with each log record hash (:timestamp, :severity_number,
+      #   :severity_text, :body, :attributes) before it is sent to PostHog Logs. Return a (possibly
+      #   modified) hash to send, or nil to drop the record — useful for scrubbing PII. If the callback
+      #   raises, the record is dropped. Defaults to nil.
+      attr_accessor :logs_before_send
+
+      # @return [Hash] Extra OpenTelemetry resource attributes merged with auto-detected service metadata.
+      attr_accessor :logs_resource_attributes
```
turnipdabeets: I'd cut logs_resource_attributes for now since it exposes a raw OTel concept as public config and based on our JS SDK we're heading toward first-class logs support without depending on OTel

posthog-rails/lib/posthog/rails/configuration.rb
```diff
@@ -33,6 +33,30 @@ class Configuration
       #   posthog_distinct_id, distinct_id, id, pk, uuid in order.
       attr_accessor :user_id_method
 
+      # @return [Boolean] Master switch for forwarding logs to PostHog Logs over OTLP. Defaults to false.
+      attr_accessor :logs_enabled
+
+      # @return [Boolean] Whether to broadcast Rails.logger output into the PostHog Logs sink. Defaults to true
+      #   (only takes effect when {#logs_enabled} is true).
+      attr_accessor :forward_rails_logger
```
turnipdabeets: Tiny nit since everything else has a `logs_` prefix could we rename this to `logs_forward_rails_logger`?

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -118,12 +122,20 @@ def ensure_initialized!
...
+
       # Ensure PostHog shuts down gracefully (register only once)
       config.after_initialize do
         next if @posthog_at_exit_registered
 
         @posthog_at_exit_registered = true
-        at_exit { PostHog.client&.shutdown if PostHog.initialized? }
+        at_exit do
+          PostHog::Rails::Logs::Setup.shutdown!
```
turnipdabeets: `provider.shutdown` here has no timeout, and it force-flushes through the exporter's retry/backoff. During a PostHog or network outage that can eat the whole SIGTERM grace period, so the orchestrator SIGKILLs the process before the *events* client on the next line ever gets to flush — losing both. Passing a small explicit `timeout:` (a second or two) bounds the damage.

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -144,6 +156,41 @@ def insert_middleware_before(app, target, middleware)
         app.config.middleware.insert_before(target, middleware)
       end
 
+      # Build the PostHog Logs pipeline and broadcast Rails.logger into it.
+      #
+      # @api private
+      # @return [void]
+      def self.install_posthog_logs
+        return unless PostHog.initialized?
```
turnipdabeets: If someone sets `logs_enabled = true` but `PostHog.init` never ran (conditional initializer, lazy init), this returns silently and they get no logs and no breadcrumb. A warn here would save someone a confusing afternoon.

posthog-rails/README.md
```diff
@@ -7,3 +7,40 @@ For installation, configuration, usage, and troubleshooting, see the official do
...
+
+`posthog-rails` can forward `Rails.logger` output to [PostHog Logs](https://posthog.com/docs/logs)
+over OpenTelemetry (OTLP), automatically correlated with the request's PostHog
+distinct ID and session ID.
+
+This is opt-in and relies on the standard OpenTelemetry gems (Ruby 3.3+), which
+are not bundled. Add them to your `Gemfile`:
+
+```ruby
+gem 'opentelemetry-sdk'
```
turnipdabeets: Worth adding `require: false` to all three gems (here, the example initializer, and the generator template/output). Without it, `Bundler.require` loads them at boot even when `logs_enabled` is false — which defeats the nice lazy-require design in `Setup` — and `opentelemetry-logs-sdk` patches `OpenTelemetry::SDK::Configurator` at require time, so an app already using OTel *tracing* silently grows a second logs pipeline pointed at `OTEL_EXPORTER_OTLP_ENDPOINT` (localhost:4318 by default). `Setup.require_otel_gems` already requires them explicitly when needed.

posthog-rails/examples/posthog.rb
turnipdabeets: Perfect, I agree we should drop rather than sending the log unmodified. Thank you!

posthog-rails/lib/posthog/rails/logs/appender.rb
turnipdabeets: ah ok yes, desired state but probably worth it's own PR.

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+        #
+        # Unlike the events before_send (which sends the original event when the
+        # callback raises), a failing callback drops the record: the likeliest
+        # use is PII scrubbing, where shipping the unscrubbed original is worse
+        # than losing the line.
+        def apply_before_send(record)
+          return record unless @before_send
+
+          result = @before_send.call(record)
+          result.is_a?(Hash) ? result : nil
```
turnipdabeets:
```suggestion
          return result if result.is_a?(Hash)
          warn_before_send_error("returned #{result.class} instead of a Hash or nil") unless result.nil?
          nil
```

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+
+          @before_send_error_warned = true
+          PostHog::Logging.logger.warn(
+            "logs_before_send raised (#{error.class}: #{error.message}); dropping records that fail the callback"
+          )
+        end
+
+        def body_for(message)
+          message.is_a?(String) ? message : message.inspect
+        end
```
turnipdabeets, line 211: dup so logs_before_send can't mutate (or trip over frozen) app strings, and UTF-8 normalization because the OTLP exporter sanitizes attribute encoding but not the body, because one bad body silently drops the whole batch.
```suggestion
        def body_for(message)
          str = message.is_a?(String) ? message.dup : message.inspect
          unless str.encoding == Encoding::UTF_8
              str = str.encode(Encoding::UTF_8, invalid: :replace, undef: :replace)
          end
          str.valid_encoding? ? str : str.scrub
        end
```

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,225 @@
...
+          # US cloud endpoint.
+          def resolve_host
+            normalize(@client_host) ||
+              normalize(ENV.fetch('POSTHOG_HOST', nil)) ||
+              'https://us.i.posthog.com'
+          end
+
+          def build_rate_limiter(config)
+            limit = config.logs_max_records_per_minute
+            return nil unless limit.is_a?(Numeric) && limit.positive?
```
turnipdabeets: ENV['LOGS_CAP'] is a String which fails the Numeric check, and silently disables the cap. Also worth a doc note that 0/negative means "no cap"

posthog-rails/lib/posthog/rails/logs/setup.rb
turnipdabeets: sounds good, thank you!

posthog-rails/lib/posthog/rails/railtie.rb
turnipdabeets: always good to have more tests :)

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+      # BatchLogRecordProcessor synchronizes its buffer internally — the same
+      # split as stdlib `Logger`, which locks in `LogDevice`, not
+      # `Logger#add`. A mutex around emit would serialize all app logging
+      # needlessly.
+      #
+      # @api private
+      class Appender < ::Logger
+        SELF_LOG_PREFIX = '[posthog-ruby]'
+        SELF_LOG_PROGNAME = 'PostHog'
+        REQUEST_ATTRIBUTE_KEYS = %w[$current_url $request_method $request_path].freeze
```
turnipdabeets: web SDK uses url.full so I think we want to match that here instead of $current_url

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+            else
+              message = progname
+              progname = nil
+            end
+          end
+
+          return true if message.nil?
+          return true if self_log?(message, progname)
+
+          case @rate_limiter&.record
```
turnipdabeets: Ordering is flipped vs every other SDK: the cross-SDK spec has beforeSend run *before* the rate cap, this way records dropped by beforeSend don't consume the window budget

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+        end
+
+        private
+
+        def emit(severity, message, progname)
+          severity_number, severity_text = Severity.for(severity)
+          record = {
+            timestamp: Time.now,
+            severity_number: severity_number,
+            severity_text: severity_text,
```
turnipdabeets: Could the callback hash expose `severity` instead of `severity_number` / `severity_text`? The other SDKs map an enum like `warn` to Otel number and text 13 / 'WARN'.

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+            body: body_for(message),
+            attributes: attributes_for(progname)
+          }
+          record = apply_before_send(record)
+          return if record.nil?
+
+          @otel_logger.on_emit(**record)
```
turnipdabeets: Can we include `traceId`/`spanId`? The other SDKs' log records carry the trace fields, and the OTel logs SDK's `on_emit` should resolve the current active span from ambient context (or an explicit `context:` argument) — so apps already running OTel tracing may get logs↔trace correlation for free here. Worth verifying against our gem versions: if it already works; if not, it's likely a one-line `context:` argument.

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,171 @@
...
+          )
+        end
+
+        def body_for(message)
+          message.is_a?(String) ? message : message.inspect
+        end
+
+        def attributes_for(progname)
+          attributes = {}
+          attributes['logger.progname'] = progname.to_s if progname
```
turnipdabeets: Let's use `logger.name` vs`logger.progname`? It's the OTel-world convention for "which logger emitted this" (what the Java/Python log bridges use), while progname is a Ruby-ism.

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,225 @@
...
+            attrs = {
+              'service.name' => service_name,
+              'deployment.environment' => ::Rails.env.to_s
+            }
+            attrs.merge(stringify_keys(config.logs_resource_attributes || {}))
+          end
+
+          def service_name
+            app = ::Rails.application
+            return 'rails' unless app
```
turnipdabeets: Fall back to `'unknown_service'` instead of `'rails'`? That's what the other SDKs use (Android/JS/RN) and the OTel-prescribed default. `service.name` is lifted into its own column server-side and users filter on it — `'rails'` looks like a real service name, while `unknown_service` correctly signals "derivation failed, fix your setup." Same on the rescue path below.

turnipdabeets:
PR is looking really good. We'll also need a .changeset file for release automation system to run

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Okay so, this "just works", I verified it against the source: `on_emit` defaults `context: OpenTelemetry::Context.current` and fills the top-level `trace_id`/`span_id`/`trace_flags` from the active span (https://github.com/open-telemetry/opentelemetry-ruby/blob/main/logs_sdk/lib/opentelemetry/sdk/logs/logger.rb). Apps that run the tracing instrumentation gems (`opentelemetry-instrumentation-rails`/`-rack`) get logs↔trace correlation for free; since our appender emits synchronously on the calling thread, the ambient context is the traced request's own.

My smoke test has no tracer, so the trace fields are empty as expected:

Screen Shot 2026-06-11 at 8 34 13 PM


I did add a line to the README documenting that traced requests get `trace_id`/`span_id` on their log records automatically.

Thoughts?

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Agreed - the callback hash now carries a single severity symbol (:debug/:info/:warn/:error/:fatal); the OTel number/text pair is derived internally after the callback runs, same as the other SDKs.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Ah good call. I flipped the order - beforeSend runs first, only surviving records consume window budget, and the cap notice is now emitted directly so a scrubber can't accidentally suppress it. Specs cover both.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Done. I went ahead and changed all three request attributes to now use OTel semconv names matching the web SDK: url.full, http.request.method, url.path.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: good catch. i added some specs.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: I took the suggestion with one adjustment: warn_before_send_error expected an exception (it formats class/message), so I generalized it to take a description and routed both the raised and wrong-type paths through it. Specs coverage

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Fixed - numeric strings are now coerced (Integer(raw, exception: false)), and unparseable values warn and fall back to the default cap rather than disabling. nil/0/negative are the documented intentional disables, noted in the config docs, initializer template, and README.

posthog-rails/README.md
johnnagro: Good call. Done everywhere the gems are mentioned (README, generator output, initializer template/example, and the LoadError warning).

posthog-rails/lib/posthog/rails/railtie.rb
johnnagro: good idea. now warns when logs_enabled is set but PostHog.init never ran.

posthog-rails/lib/posthog/rails/railtie.rb
johnnagro: Fixed - provider.shutdown(timeout: 2). Two seconds drains a healthy buffer easily and bounds the damage when the endpoint is down. force_flush also takes an optional timeout now for parity.

posthog-rails/lib/posthog/rails/configuration.rb
johnnagro: Done - renamed to logs_forward_rails_logger everywhere (config, railtie, template/example, specs).

posthog-rails/lib/posthog/rails/configuration.rb
johnnagro: Sounds good, I removed logs_resource_attributes entirely.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Good catch - no callers, and shutdown! already flushes buffered records via the provider's shutdown. Removed.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: yeah that fallback was too quiet. resolve_level now warn_onces with the bad value and the valid symbols (e.g. Invalid logs_level :warning; expected one of :debug, :info, :warn, ...) before falling back to the Rails logger level, matching how the other misconfigs in this file speak up. Added specs for both paths.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Confirmed both behaviors against the 7.1+ BroadcastLogger source (debug? is any? over sinks, level= dispatches to all). One wrinkle with leaving Logger#level alone: stdlib Logger defaults to DEBUG, so the appender would report debug? == true and widen the broadcast unconditionally. Instead the threshold lives in a private ivar checked in #add, and the inherited level is pinned to UNKNOWN so this sink is invisible to the broadcast's predicates/min — and a broadcast-wide level= can no longer clobber logs_level. Net contract (now in the README): logs_level filters what's forwarded, it never changes what the app logs. Also documented the silence and tagged/log_tags gaps. Added specs against a real BroadcastLogger for both fixes.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Both good catches - took your fix nearly as written. #add now has a fiber-local re-entrancy guard (nested records from a callback logging through the broadcast are dropped instead of recursing to SystemStackError), and the trailing rescue warns once with the error class/message before going quiet, matching warn_before_send. The guard doubles as loop protection for that warning itself. Added specs for both: a before_send that logs through a real BroadcastLogger containing the appender, and a persistently-raising on_emit.

posthog-rails/lib/posthog/rails/railtie.rb
johnnagro: Added a fork-safety spec that runs the real BatchLogRecordProcessor with a pipe-backed exporter: log in the parent to start the worker thread, fork, log in the child, and assert the child's record is exported. It pins the reset_on_fork behavior we're relying on, so an OTel SDK regression (or a swap to a processor without fork detection) shows up in CI instead of as silently unflushed worker logs.

johnnagro:
thanks, @turnipdabeets - please take another look

posthog-rails/lib/posthog/rails/logs/severity.rb
```diff
@@ -0,0 +1,55 @@
...
+      #
+      # @api private
+      module Severity
+        module_function
+
+        # @param severity [Integer, nil] A Ruby `Logger` severity constant.
+        # @return [Array(Integer, String)] OpenTelemetry severity number and text.
+        def for(severity)
+          for_name(name_for(severity))
+        end
```
turnipdabeets: this might be dead code now that we can remove?

posthog-rails/lib/posthog/rails/logs/appender.rb
turnipdabeets: I think as it's document that the tracing instrumentation gems are needed it should be ok. Ideally down the road we'd populate trace context ourselves so no extra gems are needed.

turnipdabeets:
@johnnagro "Commits must have verified signatures" in order to merge - mind signing these?

posthog-rails/lib/posthog/rails/logs/severity.rb
johnnagro: Yep - leftover from before we split the path for logs_before_send (symbol in the callback hash, OTel pair derived at emit). Removed.

johnnagro:
> @johnnagro "Commits must have verified signatures" in order to merge - mind signing these?

np - signed.

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,238 @@
...
+          #
+          # @api private
+          # @param options [Hash] The options passed to {PostHog::Client.new}.
+          # @return [void]
+          def remember_client_options(options)
+            return unless options.is_a?(Hash)
+
+            @client_api_key = options[:api_key] || options['api_key']
+            @client_host = options[:host] || options['host']
+          end
```
dustinbyrne, line 99: We should also consider `test_mode` here so that we don't emit logs in test

https://github.com/PostHog/posthog-ruby/blob/0ab66925011c53827dc8f7496babc8d4dc102413/posthog-rails/lib/generators/posthog/templates/posthog.rb#L111-L112

posthog-rails/lib/posthog/rails/logs/appender.rb
```diff
@@ -0,0 +1,245 @@
...
+            "PostHog Logs failed to emit a record (#{error.class}: #{error.message}); " \
+            'further failures will be dropped silently'
+          )
+        end
+
+        def body_for(message)
+          str = message.is_a?(String) ? message.dup : message.inspect
+          str = str.encode(Encoding::UTF_8, invalid: :replace, undef: :replace) unless str.encoding == Encoding::UTF_8
+          str.valid_encoding? ? str : str.scrub
+        end
```
dustinbyrne, line 211: If logging an exception, e.g.:
```rb
Rails.logger.error(ex)
```

It'll render as something like `<#RuntimeError: unknown error occurred>` instead of including the backtrace

we should consider doing something like [`Logger::Formatter#msg2str`](https://github.com/ruby/logger/blob/4d414691b28c02a7247023746955fba98980a316/lib/logger/formatter.rb#L25-L33)

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,238 @@
...
+          #
+          # Idempotent: subsequent calls return the previously built appender
+          # (or nil if setup was skipped).
+          #
+          # @return [PostHog::Rails::Logs::Appender, nil]
+          def install!
+            return @appender if @installed
+
+            @installed = true
+            return nil unless require_otel_gems
```
dustinbyrne: Worth checking here if the PostHog client is disabled and skipping installation if it is

We have this pattern in a lot of methods internal to the client:
https://github.com/PostHog/posthog-ruby/blob/0ab66925011c53827dc8f7496babc8d4dc102413/lib/posthog/client.rb#L293-L294

https://github.com/PostHog/posthog-ruby/blob/0ab66925011c53827dc8f7496babc8d4dc102413/lib/posthog/client.rb#L208-L209

`@disabled` isn't exposed, but I think it's fine to expose it as a public method. Maybe we opt for the positive case, though - `client.enabled?`

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,238 @@
...
+          rescue StandardError => e
+            warn_once("Failed to initialize PostHog Logs: #{e.message}")
+            nil
+          end
+
+          # Shut the pipeline down, flushing buffered records.
+          #
+          # @param timeout [Numeric] Max seconds to spend; see {SHUTDOWN_TIMEOUT_SECONDS}.
+          # @return [void]
+          def shutdown!(timeout: SHUTDOWN_TIMEOUT_SECONDS)
```
dustinbyrne: nit considering this is private API space, but I think we should prefer a non-bang method name here (just `shutdown`)

it doesn't have a non-bang form, and it's not at risk of raising an exception

`install!` maybe has a stronger case given it's highly state changing, but I think we could do a non-bang there too

Gemfile
```diff
@@ -15,6 +10,9 @@ group :development, :test do
   gem 'activesupport', '~> 7.1'
   gem 'commander', '~> 5.0'
   gem 'oj', '~> 3.16.10'
+  # Soft dependency of posthog-rails' PostHog Logs feature; present here only
+  # so the fork-safety spec can exercise the real BatchLogRecordProcessor.
+  gem 'opentelemetry-logs-sdk', require: false
```
dustinbyrne: I tried running this as is, and got the following error
```text
OpenTelemetry error: unexpected error in OTLP::Exporter#encode -
undefined method 'event_name' for an instance of OpenTelemetry::SDK::Logs::LogRecordData
```

My lockfile resolved to:
```text
  opentelemetry-logs-sdk              0.4.0
  opentelemetry-exporter-otlp-logs    0.5.1
```

and it looks like the exporter expects `log_record_data.event_name` to exist, but `opentelemetry-logs-sdk` v0.4.0 defines `LogRecordData` without an `event_name` field.

a fresh bundle with the following fixed it:
```rb
  gem 'opentelemetry-sdk', require: false                # resolved to 0.6.0
  gem 'opentelemetry-logs-sdk', require: false
  gem 'opentelemetry-exporter-otlp-logs', require: false # resolved to 0.5.1
```

maybe we use something like this?
```suggestion
  gem 'opentelemetry-logs-sdk', '>= 0.6.0', require: false
  gem 'opentelemetry-exporter-otlp-logs', require: false
```

dustinbyrne: might be worth a test using real exports from these libs

dustinbyrne (commented):
Hey @johnnagro, thanks for this! I took a look and have a few comments, but otherwise this is looking pretty solid

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Good idea, done. remember_client_options now captures test_mode and install! skips the pipeline (quietly, since it's intentional) when it's on, mirroring the core client swapping in a NoopWorker. So an app that sets config.test_mode = true if Rails.env.test? won't ship real log records from its suite. Added a spec.

posthog-rails/lib/posthog/rails/logs/appender.rb
johnnagro: Ah, good catch. body_for now mirrors Logger::Formatter#msg2str, so PostHog matches the file log. Guarded the nil-backtrace case for never-raised exceptions and kept our UTF-8 safety on top. Added specs.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Done - added a public PostHog::Client#enabled? and gate install_posthog_logs on it, skipping quietly. if you decide you'd rather not alter the core client with this change, we could alternatively stash this in our init options like we did with api_key

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: sure, neither has a non-bang form and both rescue internally, so the bang was misleading. Renamed install!→install and shutdown!→shutdown. Left reset! as-is since the bang reads as a meaningful "throws away all state" for a test helper, but happy to drop it too if you'd rather be uniform.

Gemfile
johnnagro: Confirmed and fixed. The OTLP exporter (≥0.5.0) encodes LogRecordData#event_name, which only exists in logs-sdk ≥0.6.0, so our lock's 0.4.0 broke real exports - the fork spec missed it because it used a fake exporter. Pinned opentelemetry-logs-sdk >= 0.6.0 (+ added the exporter) in the dev Gemfile, and added the >= 0.6.0 floor to the README/generator/templates so users don't hit it. Note 0.6.0 needs Ruby 3.3+, so the dev gems are gated to 3.3+ (matching the feature's documented requirement) and the logs specs skip on the 3.2 CI leg. Also took your second suggestion: a real-export spec that stubs the endpoint with WebMock and asserts the actual protobuf POST goes out, so the encode path is covered and this class of mismatch fails loudly.

posthog-rails/lib/posthog/rails/logs/setup.rb
turnipdabeets: I think it would be better to not modify the public API here, wdyt @dustinbyrne ?

dustinbyrne: I'm fine with adding `enabled?`, otherwise we have to use reflection or duplicate the `@disabled` logic (which, admittedly is minimal, but at least we're resilient to it changing in the future).

dustinbyrne:
Thank for the changes @johnnagro! I don't see a new commit yet, so I'll circle back on this

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -0,0 +1,238 @@
...
+          def logs_endpoint(host)
+            base = (host || 'https://us.i.posthog.com').to_s.sub(%r{/+\z}, '')
+            "#{base}/i/v1/logs"
+          end
+
+          def resolve_level(level)
+            return nil if level.nil?
+            return level if level.is_a?(Integer)
+
+            ::Logger.const_get(level.to_s.upcase)
```
turnipdabeets: nit, nice to have
```suggestion
            result = ::Logger.const_get(level.to_s.upcase)
    		raise NameError, "not a severity integer" unless result.is_a?(Integer)
    		result
```

posthog-rails/lib/posthog/rails/logs/setup.rb
dustinbyrne: Thanks! I missed `reset!` - would prefer no bang there but also not a blocker given it's not a public API

johnnagro: Done. reset! → reset. Now consistent with install/shutdown.

posthog-rails/lib/posthog/rails/logs/setup.rb
johnnagro: Good catch. const_get can resolve non-severity constants (Logger::VERSION, or inherited ENV/ARGV) without raising, breaking the comparison later. Added the Integer guard to reject them; specs added.

johnnagro:
@turnipdabeets @dustinbyrne Thanks for the review & comments. I believe I've addressed all outstanding feedback.

turnipdabeets:
> @turnipdabeets @dustinbyrne Thanks for the review & comments. I believe I've addressed all outstanding feedback.

LGTM, @dustinbyrne want to do one more pass at the changes you requested? 

Actually might have spoken too soon, looks like we have some tests failing now. 

@johnnagro looks like we might need the Gemfile.lock committed to fix CI.

dustinbyrne:
looks good, we can approve once the bundle is updated and CI passes 👍 

thanks again @johnnagro

johnnagro:
@turnipdabeets @dustinbyrne ah okay - let me dig in. this relates to the OTel dependencies not being compatible with 3.2, i was trying to make them conditional on ruby version. let me dig in.

johnnagro:
@turnipdabeets @dustinbyrne I pulled the OTel gems back out of the Gemfile. Bundling them broke the Ruby 3.2 leg. The fork-safety and real-export specs stay but are guarded, so they skip in CI when the gems aren't present and still run locally on 3.3+. The >=0.6.0 requirement is documented for users. If you'd like the real-export spec to run in CI too, I can add a dedicated gemfiles/otel.gemfile + a 3.3/3.4-only job, which keeps frozen mode and the cooldown/checksums intact - happy to do that as a follow-up if you want it.

johnnagro:
I’ll fix the public api check

johnnagro:
@turnipdabeets @dustinbyrne i believe i've regenerated the public api snapshot.

turnipdabeets (approved):
Thanks @johnnagro for your contribution!

PR #176, opened 2026-06-17
feat: add flush interval batching

marandaneto:

Ruby async batching previously sent whatever was queued as soon as the worker ran. Other server SDKs (Python, .NET, Go, Elixir) support a time-based flush threshold so partial batches can accumulate briefly instead of becoming one request per event under low throughput.

This adds a configurable `flush_interval_seconds` for async batching, defaulting to 5 seconds, aligned with the event-batcher spec and the Python SDK behavior from PostHog/posthog-python#672.

- `bundle exec rubocop lib/posthog/send_worker.rb lib/posthog/client.rb lib/posthog/noop_worker.rb spec/posthog/send_worker_spec.rb`
- `bundle exec rspec`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi/coding-agent. I used the existing Ruby worker/client design and referenced batching behavior from the Python SDK and the shared event-batcher spec. The implementation keeps explicit `flush` and shutdown paths immediate while allowing normal async workers to wait up to the configured interval for a fuller batch.


spec/posthog/send_worker_spec.rb
```diff
@@ -101,10 +109,68 @@ def bad_obj.to_json(*_args)
...
+        queue << Requested::CAPTURE
+        worker = described_class.new(queue, 'testsecret', batch_size: 10, flush_interval: 60)
+
+        worker_thread = Thread.new { worker.run }
+        eventually { expect(worker.is_requesting?).to eq(true) }
+        worker.request_flush
+
+        expect(worker_thread.join(1)).to eq(worker_thread)
+        expect(sends).to eq([1])
+      end
```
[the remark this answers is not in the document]
marandaneto: Added coverage in 8479aee for the producer wake path: the test now starts a worker with one queued event, enqueues a second event while the worker is waiting, calls `notify`, and asserts both events are sent together before the interval elapses.

ioannisj (commented):
Got this from my bot, may worth checking if this is a valid concern?

```
Pre-PR flush() removed the batch before sending:
$batch = array_splice($this->queue, 0, min($this->batch_size, $count)); // REMOVE first
$success = $this->flushBatch($batch);                                    // then send
On a socket-open failure, old flushBatch() returned false → the batch was already gone (dropped/lost) and the loop stopped. The queue never grew.

Post-PR flush() copies, sends, then conditionally removes (QueueConsumer.php:119-130):
$batch  = array_slice($this->queue, 0, $batchSize);          // COPY, don't remove
$result = $this->flushBatch($batch);
$success = true === $result;
if ($success || self::FLUSH_BATCH_RETRYABLE_FAILURE !== $result) {
    array_splice($this->queue, 0, $batchSize);               // remove only if NOT retryable
}
Now a socket-open failure returns FLUSH_BATCH_RETRYABLE_FAILURE (Socket.php:52-53, changed from return false by this PR) → the batch is retained and the loop stops
```

marandaneto:
I checked this concern and it doesn't apply to this Ruby PR. The referenced files/symbols (`QueueConsumer.php`, `Socket.php`, `FLUSH_BATCH_RETRYABLE_FAILURE`, `array_slice`/`array_splice`) are from the PHP SDK and don't exist in this repository. The Ruby worker still removes events from the queue before sending by moving them into `MessageBatch` via `Queue#pop`, and this PR doesn't add retry-retention semantics or change failure retry behavior. No code change needed here.

marandaneto:
> I checked this concern and it doesn't apply to this Ruby PR. The referenced files/symbols (`QueueConsumer.php`, `Socket.php`, `FLUSH_BATCH_RETRYABLE_FAILURE`, `array_slice`/`array_splice`) are from the PHP SDK and don't exist in this repository. The Ruby worker still removes events from the queue before sending by moving them into `MessageBatch` via `Queue#pop`, and this PR doesn't add retry-retention semantics or change failure retry behavior. No code change needed here.

this PR is ruby and not PHP

lib/posthog/send_worker.rb
```diff
@@ -82,20 +103,81 @@ def is_requesting? # rubocop:disable Naming/PredicateName
...
     end
 
+    def wait_for_work
+      @state_lock.synchronize do
+        while @queue.empty? && !@shutdown
+          clear_flush_request_without_lock
+          @condition.wait(@state_lock)
+        end
+      end
+    end
```
dustinbyrne, line 169: edge case - forking while the async worker has a partial batch can deadlock flush

not 100% sure how feasible it is to trigger in practice, i think it's unlikely to trigger via puma or unicorn (maybe a capture during init?), but something like resque could fork at any time

```rb
it 'completes after fork when the async worker has a partial batch' do
  PostHog::Transport.stub = true
  async_client = Client.new(api_key: API_KEY, batch_size: 10, flush_interval_seconds: 60)
  worker = async_client.instance_variable_get(:@worker)
  queue = async_client.instance_variable_get(:@queue)

  async_client.capture(Queued::CAPTURE)
  eventually do
    expect(queue).to be_empty
    expect(worker.is_requesting?).to eq(true)
  end

  reader, writer = IO.pipe
  pid = Process.fork do
    reader.close
    flush_thread = Thread.new { async_client.flush }
    result = flush_thread.join(0.5) == flush_thread ? 'returned' : 'hung'
    flush_thread.kill if flush_thread.alive?
    writer.write(result)
    writer.close
    exit! 0
  end

  writer.close
  result = reader.read
  Process.wait(pid)
  reader.close

  expect(result).to eq('returned')
ensure
  async_client&.shutdown
  PostHog::Transport.stub = false
end
```

marandaneto: Good catch, this was valid. Fixed in 90da694 by making SendWorker detect forked processes, clear inherited in-flight batch state, reset shutdown/flush flags, and recreate the transport in the child. I also added the fork regression test for `flush` while the async worker has a partial batch.

ioannisj (approved):
LGTM

PR #179, opened 2026-06-17
feat(rails): add current user resolver

### LINKED ISSUE #102: Improvements to user tracking
Right now it's quite limited with `current_user_method`. Ideally we could pass in a block to execute, because "the modern Rails way" is to use `Current` so I'd expect most newer apps would have current user defined on `Current.user`.

marandaneto:

Closes #102.

Rails apps that use `ActiveSupport::CurrentAttributes` (for example, `Current.user`) currently need to define a controller method solely for PostHog exception user context. This adds a resolver hook so apps can provide the current user directly while keeping the existing `current_user_method` behavior.

- `BUNDLE_PATH=../posthog-ruby/vendor/bundle bundle exec rspec spec/posthog/rails/request_context_spec.rb spec/posthog/rails/configuration_spec.rb`
- `BUNDLE_PATH=../posthog-ruby/vendor/bundle bundle exec rubocop posthog-rails/lib/posthog/rails/capture_exceptions.rb spec/posthog/rails/request_context_spec.rb`
- `BUNDLE_PATH=../posthog-ruby/vendor/bundle bundle exec rake public_api:check`
- `BUNDLE_PATH=../posthog-ruby/vendor/bundle bundle exec rspec`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi coding agent. The change adds `PostHog::Rails.config.current_user_resolver`, which takes precedence over `current_user_method` and supports both no-argument resolvers like `proc { Current.user }` and controller-aware resolvers. Existing controller method behavior is preserved for backwards compatibility. Controller-aware resolvers safely no-op when an exception occurs before Rails has instantiated a controller, and resolver failures fall back to request/tracing context instead of dropping exception capture.

### DIFF
--- a/posthog-rails/README.md
+++ b/posthog-rails/README.md
@@ -9,20 +9,6 @@ SDK usage examples and code snippets live in the official documentation so they 
 - [Ruby on Rails framework docs](https://posthog.com/docs/libraries/ruby-on-rails)
 - [Ruby library docs](https://posthog.com/docs/libraries/ruby)
 
-## User context for exception capture
-
-By default, exception capture calls `current_user` on the controller. Apps that
-use `ActiveSupport::CurrentAttributes` can provide a resolver instead:
-
-```ruby
-PostHog::Rails.configure do |config|
-  config.current_user_resolver = proc { Current.user }
-end
-```
-
-The resolver takes precedence over `current_user_method` and may accept the
-controller as an argument.
-
 ## PostHog Logs (optional)
 
 `posthog-rails` can forward `Rails.logger` output to [PostHog Logs](https://posthog.com/docs/logs)
--- a/posthog-rails/README.md
+++ b/posthog-rails/README.md
@@ -9,6 +9,20 @@ SDK usage examples and code snippets live in the official documentation so they 
 - [Ruby on Rails framework docs](https://posthog.com/docs/libraries/ruby-on-rails)
 - [Ruby library docs](https://posthog.com/docs/libraries/ruby)
 
+## User context for exception capture
+
+By default, exception capture calls `current_user` on the controller. Apps that
+use `ActiveSupport::CurrentAttributes` can provide a resolver instead:
+
+```ruby
+PostHog::Rails.configure do |config|
+  config.current_user_resolver = proc { Current.user }
+end
+```
+
+The resolver takes precedence over `current_user_method` and may accept the
+controller as an argument.
+
 ## PostHog Logs (optional)
 
 `posthog-rails` can forward `Rails.logger` output to [PostHog Logs](https://posthog.com/docs/logs)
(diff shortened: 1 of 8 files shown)


posthog-rails/README.md
```diff
@@ -9,6 +9,20 @@ SDK usage examples and code snippets live in the official documentation so they
 - [Ruby on Rails framework docs](https://posthog.com/docs/libraries/ruby-on-rails)
 - [Ruby library docs](https://posthog.com/docs/libraries/ruby)
 
+## User context for exception capture
```
marandaneto: rollback this file

marandaneto: Rolled back the README change in 8f1b1e0.

posthog-rails/lib/posthog/rails/capture_exceptions.rb
```diff
@@ -69,20 +69,32 @@ def capture_exception(exception, env)
...
+
+        controller.send(method_name)
+      end
+
+      def call_current_user_resolver(resolver, controller)
+        return controller.instance_exec(&resolver) if resolver.arity.zero? && controller
+        return resolver.call if resolver.arity.zero?
+
+        resolver.call(controller)
+      end
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 064b26b — controller-aware resolvers now no-op when no controller is available, preserving exception capture fallback context.

spec/posthog/rails/request_context_spec.rb
```diff
@@ -227,6 +227,82 @@ def current_user
...
+    end
+    middleware = described_class.new(PostHog::Rails::CaptureExceptions.new(app))
+
+    expect do
+      middleware.call(env_for('/boom'))
+    end.to raise_error(StandardError, 'boom')
+
+    message = client.dequeue_last_message
+    expect(message[:distinct_id]).to eq('resolved-user')
+  end
```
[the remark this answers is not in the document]
marandaneto: Addressed in 064b26b — consolidated the resolver specs into a parameterized example and added coverage for a controller-aware resolver when no controller exists.

posthog-rails/lib/posthog/rails/capture_exceptions.rb
```diff
@@ -69,20 +69,33 @@ def capture_exception(exception, env)
...
 
-          if controller.respond_to?(method_name, true)
-            user = controller.send(method_name)
-            user_id = extract_user_id(user) if user
-            return user_id if present?(user_id)
-          end
-        end
+        user = extract_current_user(env['action_controller.instance'])
+        user_id = extract_user_id(user) if user
+        return user_id if present?(user_id)
```
ioannisj, line 76: we probably want to return nil from this method if the user provided user resolver raises. Otherwise the whole exception is dropped because of a bad provider?

marandaneto: Good point — fixed in e8028db. Resolver failures now return nil (with a warning) so exception capture falls back to request/tracing context instead of being dropped.

dustinbyrne (approved):
nice, good change 👍

PR #181, opened 2026-06-17
fix(rails): handle missing middleware insertion targets

### LINKED ISSUE #99: middleware insertion limitation - If `target` doesn't exist then we won't insert our middleware
There's a problem here, though. If `target` doesn't exist then we won't insert our middleware, and that's a bad failure. I wonder if there's a better way. Will keep it like this for now, but I'll keep thinking about it.

_Originally posted by @rafaeelaudibert in https://github.com/PostHog/posthog-ruby/pull/98#discussion_r2779403700_

marandaneto:

Fixes #99.

Rails applications can customize their middleware stack so `ActionDispatch::ShowExceptions` or `ActionDispatch::DebugExceptions` may be absent. The Railtie previously recorded direct `insert_before` / `insert_after` calls, which could fail later when Rails built the real middleware stack.

This change records fallback-aware middleware operations for Rails' deferred middleware proxy. If the target middleware is missing, PostHog still inserts its middleware and logs a warning instead of failing or silently skipping it.

- `BUNDLE_PATH=/Users/marandaneto/Github/posthog-ruby/vendor/bundle bundle exec rubocop posthog-rails/lib/posthog/rails/railtie.rb spec/posthog/rails/railtie_spec.rb`
- `BUNDLE_PATH=/Users/marandaneto/Github/posthog-ruby/vendor/bundle bundle exec rspec spec/posthog/rails`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi in a dedicated worktree. The fix keeps the existing Railtie API but replaces direct deferred middleware proxy calls with fallback-aware operations, so normal insertion order is preserved when targets exist and safe insertion is used when targets are absent.

### DIFF
--- a/posthog-rails/lib/posthog/rails/railtie.rb
+++ b/posthog-rails/lib/posthog/rails/railtie.rb
@@ -147,23 +147,88 @@ module PostHog
         end
       end
 
+      MISSING_MIDDLEWARE_MESSAGE = 'No such middleware to insert'
+
       # @api private
       # @return [void]
       def insert_middleware_after(app, target, middleware)
-        # During initialization, app.config.middleware is a MiddlewareStackProxy
-        # which only supports recording operations (insert_after, use, etc.)
-        # and does NOT support query methods like include?.
-        app.config.middleware.insert_after(target, middleware)
+        insert_middleware(app.config.middleware, :after, target, middleware)
       end
 
       # @api private
       # @return [void]
       def insert_middleware_before(app, target, middleware)
-        # During initialization, app.config.middleware is a MiddlewareStackProxy
-        # which only supports recording operations (insert_before, use, etc.)
-        # and does NOT support query methods like include?.
-        app.config.middleware.insert_before(target, middleware)
+        insert_middleware(app.config.middleware, :before, target, middleware)
       end
+
+      # During initialization, app.config.middleware is usually a
+      # Rails::Configuration::MiddlewareStackProxy. The proxy only records
+      # operations, so missing-target errors happen later when Rails builds the
+      # real ActionDispatch::MiddlewareStack. Add our own deferred operation so
+      # we can fall back at build time instead of crashing or silently skipping.
+      def insert_middleware(middleware_stack, location, target, middleware)
+        if middleware_stack_proxy?(middleware_stack)
+          append_middleware_operation(middleware_stack) do |resolved_stack|
+            insert_middleware_with_fallback(resolved_stack, location, target, middleware)
+          end
+        else
+          insert_middleware_with_fallback(middleware_stack, location, target, middleware)
+        end
+      end
+
+      def middleware_stack_proxy?(middleware_stack)
+        defined?(::Rails::Configuration::MiddlewareStackProxy) &&
+          middleware_stack.instance_of?(::Rails::Configuration::MiddlewareStackProxy) &&
+          middleware_stack.instance_variable_defined?(:@operations)
+      end
+
+      def append_middleware_operation(middleware_stack, &operation)
+        middleware_stack.instance_variable_get(:@operations) << operation
+      end
+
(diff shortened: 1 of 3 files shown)


posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -147,24 +147,89 @@ def ensure_initialized!
...
 
+      def middleware_stack_proxy?(middleware_stack)
+        defined?(::Rails::Configuration::MiddlewareStackProxy) &&
+          middleware_stack.instance_of?(::Rails::Configuration::MiddlewareStackProxy) &&
+          middleware_stack.instance_variable_defined?(:@operations)
+      end
+
+      def append_middleware_operation(middleware_stack, &operation)
+        middleware_stack.instance_variable_get(:@operations) << operation
+      end
```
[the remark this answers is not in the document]
marandaneto: Fixed in 93ad737 — added legacy Rails 5/6 tuple-format operation support and a regression spec so we do not append a bare Proc to older MiddlewareStackProxy operations.

spec/posthog/rails/railtie_spec.rb
```diff
@@ -98,6 +99,74 @@
...
+        MissingTargetMiddleware,
+        FallbackMiddleware
+      )
+
+      stack = ActionDispatch::MiddlewareStack.new
+      stack.use(ExistingMiddleware)
+      expect { middleware_proxy.merge_into(stack) }.not_to raise_error
+      expect(stack.middlewares.map(&:klass)).to eq([FallbackMiddleware, ExistingMiddleware])
+      expect(logger).to have_received(:warn).with(/Could not find MissingTargetMiddleware/)
+    end
```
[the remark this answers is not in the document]
marandaneto: Fixed in 93ad737 — collapsed the two fallback examples into a single parameterized example covering both before/prepend and after/append behavior.

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -147,24 +147,124 @@ def ensure_initialized!
...
+            insert_middleware_with_fallback(resolved_stack, location, target, middleware)
+          end
+        else
+          ensure_middleware_stack_fallback_operation!
+          operations << [MIDDLEWARE_FALLBACK_OPERATION, [location, target, middleware], nil]
+        end
+      end
+
+      def callable_middleware_operations?(operations)
+        return ::Rails::VERSION::MAJOR >= 7 if operations.empty?
```
ioannisj: Bot flagged this as the wrong version check here, which will likely crash the app. Mind checking as well?

posthog-rails/lib/posthog/rails/railtie.rb
```diff
@@ -147,24 +147,124 @@ def ensure_initialized!
         end
       end
 
+      MISSING_MIDDLEWARE_MESSAGE = 'No such middleware to insert'
+      MIDDLEWARE_FALLBACK_OPERATION = :posthog_insert_middleware_with_fallback
```
ioannisj, line 68: Do these change the public api?

posthog-rails/lib/posthog/rails/railtie.rb
marandaneto: Good catch, fixed in 2707774 — removed the Rails version check and now probes the MiddlewareStackProxy operation format directly before choosing callable vs legacy tuple operations.

posthog-rails/lib/posthog/rails/railtie.rb
marandaneto: Fixed in 2707774 — marked the helper constants/module as private constants and made the injected middleware-stack helper method private, so these are implementation details rather than public API.

dustinbyrne (approved):

PR #182, opened 2026-06-18
fix: Normalize ignored batch metadata fields

marandaneto:

Ruby batch payloads included top-level `library`, `library_version`, and `messageId` fields, but the batch ingestion path ignores those fields. Ingestion reads SDK metadata from event properties (`$lib`, `$lib_version`) and event identity from `uuid`.

This PR stops sending the ignored top-level fields while preserving the canonical metadata properties. Deprecated `message_id` input is normalized to `uuid` when it is a valid UUID and `uuid` is not otherwise provided. If neither `uuid` nor `message_id` is valid, the SDK now generates a UUID before enqueueing the event.

- `bundle exec rspec spec/posthog/client_spec.rb`
- `bundle exec rspec`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

An AI coding agent cross-checked the Ruby payload against the PostHog batch ingestion path, then updated the SDK to stop emitting ignored top-level metadata fields. The PR keeps SDK metadata in the canonical `$lib` / `$lib_version` properties, normalizes deprecated `message_id` to event `uuid` when possible, and generates a fresh UUID when neither provided value is valid.

Review feedback was addressed by splitting looped assertions into separately named examples per SDK action, removing the now-orphaned `Utils#uid`, and making UUID normalization fall back from an invalid `uuid` to a valid `message_id` while still warning for invalid values. A patch changeset was added for the release note.

### DIFF
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -181,9 +181,9 @@ module PostHog
     end
 
     # @!macro common_attrs
-    #   @option attrs [String] :message_id Deprecated no-op. The batch endpoint ignores `messageId`.
-    #     Use `:uuid` on capture events for event identity/deduplication instead. Read SDK metadata from
-    #     `$lib` and `$lib_version` properties, not top-level `library` / `library_version` fields.
+    #   @option attrs [String] :message_id Deprecated. Use `:uuid` instead. If `:uuid` is absent and
+    #     `:message_id` is a valid UUID, it is sent as `uuid` for backwards compatibility.
+    #     SDK metadata is sent as `$lib` and `$lib_version` properties.
     #   @option attrs [Time] :timestamp When the event occurred (optional)
     #   @option attrs [String] :distinct_id The ID for this user in your database
 
@@ -949,10 +949,6 @@ module PostHog
 
       action = process_before_send(action)
       return false if action.nil? || action.empty?
-
-      # Deprecated top-level request id retained for backwards compatibility.
-      # PostHog ingestion ignores `messageId`; use `uuid` on capture events for event identity/deduplication.
-      action[:messageId] ||= uid
 
       if @sync_mode
         send_sync(action)
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -181,9 +181,10 @@ module PostHog
     end
 
     # @!macro common_attrs
-    #   @option attrs [String] :message_id Deprecated. Use `:uuid` instead. If `:uuid` is absent and
-    #     `:message_id` is a valid UUID, it is sent as `uuid` for backwards compatibility.
-    #     SDK metadata is sent as `$lib` and `$lib_version` properties.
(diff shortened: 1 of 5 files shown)


spec/posthog/client_spec.rb
```diff
@@ -1755,10 +1753,16 @@ def length
...
-          expect(check_property.call(client.dequeue_last_message, :messageId, '5')).to eq(true)
+          message = client.dequeue_last_message
+
+          expect(message[:messageId]).to eq('5')
+          expect(message[:library]).to eq('posthog-ruby')
+          expect(message[:library_version]).to eq(PostHog::VERSION.to_s)
+          expect(message[:properties]['$lib']).to eq('posthog-ruby')
+          expect(message[:properties]['$lib_version']).to eq(PostHog::VERSION.to_s)
         end
       end
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in f6571b8 — split the looped assertions into separately named examples per action, including the sibling queue-full test.

lib/posthog/client.rb
```diff
@@ -949,9 +950,6 @@ def enqueue(action)
       action = process_before_send(action)
       return false if action.nil? || action.empty?
 
-      # add our request id for tracing purposes
-      action[:messageId] ||= uid
```
ioannisj, line 953: Robot says that with this `Utils#uid` is now orphaned, worth checking

lib/posthog/field_parser.rb
```diff
@@ -151,13 +147,13 @@ def parse_common_fields(fields)
...
         parsed = {
           timestamp: datetime_in_iso8601(timestamp),
-          library: 'posthog-ruby',
-          library_version: PostHog::VERSION.to_s,
-          messageId: message_id,
           distinct_id: distinct_id,
           properties: properties
         }
 
+        uuid = fields[:uuid] || fields[:message_id]
```
ioannisj: If `fields[:uuid]` is invalid, we drop a possibly valid `fields[:message_id]` that we could use? Unless this is what the intention is here

ioannisj: Also, invalid values are silently dropped here?

ioannisj (approved):

lib/posthog/client.rb
marandaneto: Good catch, fixed in ac24848 — removed the now-orphaned `Utils#uid` helper and its unused `securerandom` require from `utils.rb`.

lib/posthog/field_parser.rb
marandaneto: Fixed in ac24848 — UUID normalization now tries `uuid` first, then falls back to a valid `message_id` when `uuid` is absent or invalid.

marandaneto: Also covered in ac24848 — invalid `uuid` and `message_id` values still go through `valid_uuid_for_event_props?`, so they are warned and not silently accepted. Added tests for both fallback and invalid message_id handling.

PR #191, opened 2026-06-19
fix: keep changelog notices out of release notes

marandaneto:

The package changelog notices about historical pre-split entries can appear inside the latest generated release section. Because GitHub release notes are extracted from that latest changelog section, the notice also appeared in the `posthog-ruby-v3.14.1` and `posthog-rails-v3.14.1` release notes.

This keeps the changelog content as-is, but prevents the historical notice from being included in future GitHub release notes.

- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/release.yml'); puts 'workflow yaml ok'"`
- Verified the release note extraction skips the historical notice and only emits the latest release changes.

## :pencil: Checklist

- [ ] I added tests to verify the changes.

### If releasing new changes

## 🤖 Agent context

Implemented with pi. The fix leaves the changelog layout unchanged and updates the release workflow extraction logic to skip the historical package notice when creating GitHub release notes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -318,6 +318,7 @@ jobs:
           if [ -f "$CHANGELOG_FILE" ]; then
             awk -v defText="See $CHANGELOG_FILE" '
               /^## / { if (found_version) exit; found_version=1; next }
+              /^> Historical entries before the changelog split are copied from/ { next }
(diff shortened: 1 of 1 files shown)


marandaneto:
not needed anymore, fixed manually

PR #194, opened 2026-06-24
fix: Clear feature flag called cache on shutdown

marandaneto:

Feature flag call deduplication state currently remains in memory after `shutdown`, which can leak stale called-cache state across a client shutdown/reuse lifecycle. This clears the cache during shutdown while holding the existing cache mutex.

- [x] `bundle exec rspec spec/posthog/client_spec.rb`
- [x] `git diff --check`

## :pencil: Checklist

- [x] I updated the docs if needed. (Not needed.)

### If releasing new changes

- [x] Added a changeset file (`.changeset/quiet-cycles-shutdown.md`)

## 🤖 Agent context

An agent prepared this PR from the dedicated `fix/ff-called-cache-shutdown` worktree at the user's direction. The change adds a guarded shutdown-time clear of the feature flag call dedupe cache, a regression spec for that behavior, and a patch changeset for `posthog-ruby`.

Validation included the client spec file and whitespace checks. A local `pnpm changeset status --since=main` validation attempt failed because dependencies are not installed in this worktree (`changeset` executable missing), but the changeset file was added using the repository's Changesets format.

### DIFF
--- a/.changeset/quiet-cycles-shutdown.md
+++ b/.changeset/quiet-cycles-shutdown.md
@@ -1,5 +1,6 @@
 ---
 "posthog-ruby": patch
+"posthog-rails": patch
 ---
 
 Clear the feature flag call dedupe cache on shutdown.
--- a/.changeset/quiet-cycles-shutdown.md
+++ b/.changeset/quiet-cycles-shutdown.md
@@ -1,0 +1,5 @@
+---
+"posthog-ruby": patch
+---
+
+Clear the feature flag call dedupe cache on shutdown.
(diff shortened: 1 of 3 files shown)


.changeset/quiet-cycles-shutdown.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-ruby": patch
```
marandaneto, line 2: also release posthog-rails

marandaneto: Fixed in 9b37ad0 — added posthog-rails as a patch release to the changeset.

dustinbyrne (approved):

PR #197, opened 2026-06-24
feat(rails): stamp telemetry.sdk.name on forwarded logs

turnipdabeets:
## Problem

posthog-rails forwards `Rails.logger` output to PostHog Logs over OTLP, but the records carry the OpenTelemetry SDK's **default** resource attribute `telemetry.sdk.name = "opentelemetry"` (the gem never overrides it). PostHog's usage report splits per-SDK log volume by matching `telemetry.sdk.name` against the known PostHog SDKs (`posthog-ios`, `posthog-react-native`, `posthog-android`, `posthog-flutter`), so **Ruby logs match nothing and can't be attributed** — there is no `ruby_logs_records_in_period`.

## Change

Stamp `telemetry.sdk.name = "posthog-ruby"` on the OTLP resource, mirroring the convention the mobile SDKs already use.

### DIFF
--- a/posthog-rails/lib/posthog/rails/logs/setup.rb
+++ b/posthog-rails/lib/posthog/rails/logs/setup.rb
@@ -184,6 +184,11 @@ module PostHog
             # The posthog-rails name/version travel with each record via the
             # instrumentation scope (see LoggerProvider#logger above).
             {
+              # Identifies these records as the PostHog Ruby SDK so PostHog's usage
+              # report can attribute log volume to Ruby (mirrors posthog-ios/android/
+              # flutter/react-native). Overrides the OpenTelemetry SDK default of
+              # 'opentelemetry'; the posthog-rails integration name stays on the scope.
+              'telemetry.sdk.name' => 'posthog-ruby',
               'service.name' => service_name,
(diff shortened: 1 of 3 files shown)


dustinbyrne (approved):

posthog-rails/lib/posthog/rails/logs/setup.rb
```diff
@@ -184,6 +184,11 @@ def resource_attributes
             # The posthog-rails name/version travel with each record via the
             # instrumentation scope (see LoggerProvider#logger above).
             {
+              # Identifies these records as the PostHog Ruby SDK so PostHog's usage
+              # report can attribute log volume to Ruby (mirrors posthog-ios/android/
+              # flutter/react-native). Overrides the OpenTelemetry SDK default of
+              # 'opentelemetry'; the posthog-rails integration name stays on the scope.
+              'telemetry.sdk.name' => 'posthog-ruby',
```
marandaneto, line 191: @turnipdabeets we should extract this to a const and reuse since its duplicated now
see https://github.com/PostHog/posthog-ruby/blob/8dd77c89630472ddccf385e96697904220784536/lib/posthog/field_parser.rb#L147-L148

PR #206, opened 2026-07-01
ci: skip heavy CI for non-impacting changes

marandaneto:
## Summary
- add a fail-open changed-file classifier to the Unit Tests workflow
- keep required RSpec, RuboCop, Public API snapshot, and Gem build check names running on all PRs
- skip heavy steps only when every PR `filename` and `previous_filename` is in the generic safe set: markdown/docs, repository templates/CODEOWNERS/dependabot config, or the approved non-CI workflow names

## Validation
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/unit-tests.yml"); puts "YAML parsed"'
- git diff --check
- extracted classifier sample cases for safe markdown/docs/templates/CODEOWNERS/dependabot/non-CI workflows, unsafe source/test/package/build/current-CI/unknown/empty paths, and safe vs unsafe `previous_filename` rename cases
- actionlint not installed locally; skipped

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -12,11 +12,139 @@ concurrency:
 
 permissions:
   contents: read
+  pull-requests: read
 
 jobs:
+  classify-changes:
+    name: Classify changed files
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


marandaneto:
Closing this for now — I'll investigate other options. We don't want hardcoded paths this way.

PR #208, opened 2026-07-02
fix: Retry flags requests on 502 and 504

marandaneto:
## Summary
- retry remote `/flags/?v=2` evaluations on HTTP 502 and 504 using `feature_flag_request_max_retries`
- keep status-code retries opt-in to the flags evaluation request path only
- add focused specs for 502/504 retry, non-retry statuses, and max-retries opt-out

## Testing
- `ruby -c lib/posthog/feature_flags.rb`
- `bundle exec rspec spec/posthog/flags_spec.rb`
- `bundle exec rubocop lib/posthog/feature_flags.rb spec/posthog/flags_spec.rb`

### DIFF
--- a/.changeset/gentle-flags-retry.md
+++ b/.changeset/gentle-flags-retry.md
@@ -1,0 +1,5 @@
+---
+'posthog-ruby': patch
+---
+
+Retry remote feature flag requests after transient 502 and 504 responses.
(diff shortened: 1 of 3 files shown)


marandaneto:
Closing in favor of #209, which contains the same changes from a branch in the PostHog/posthog-ruby repository instead of a fork.

PR #211, opened 2026-07-02
feat: add secret_key config, deprecate personal_api_key

turnipdabeets:
## Problem

The credential used for local feature flag evaluation and remote config accepts **either** a Personal API Key (`phx_...`) **or** a Project Secret API Key (`phs_...`). Today the only option name is `:personal_api_key`, which is confusing — users with a Project Secret API Key reasonably assume the option won't accept it.

## Change

- Add a new canonical `:secret_key` option to `PostHog::Client.new` (and the Rails `config.secret_key=` setter). It accepts either a Personal API Key (`phx_...`) or a Project Secret API Key (`phs_...`).
- Keep `:personal_api_key` working as a **deprecated alias**. When only the alias is used, a deprecation warning points users at `:secret_key`.
- When both are supplied, `:secret_key` wins.
- Internally the resolved value is stored as `@secret_key`, with `@personal_api_key` kept as a mirror so existing reads (the flags poller auth headers) keep working unchanged.
- Updated YARD `@option` docs, the flag-definition-cache doc example, and the Rails/generator config examples to demonstrate the canonical option.

**Non-breaking** — existing `:personal_api_key` callers keep working.

Specs added in `spec/posthog/client_spec.rb` prove: `:secret_key` sets the credential, `:personal_api_key` still works and warns, and `:secret_key` wins when both are provided. Full suite (580 examples) passes; RuboCop clean.

## Context

- Coordinated rename across the backend SDKs (mirrors posthog-python PR #727 — this is the same change ported to Ruby).
- Rename item tracked in PostHog/posthog-js#4046 and discussed in Slack.


turnipdabeets:
**Marked ready for review.**

Naming is settled — this is a generic **`secret_key`** config that accepts **either** a Personal API Key or a Project Secret API Key (both coexist; same `Authorization: Bearer` header). `personal_api_key` stays as a deprecated **config-name** alias that forwards into it — *not* a deprecation of personal keys as a credential. It's a name-level change and is crash-safe with respect to using the key (adds no new error path).

Merge timing still coordinates with the project-secret-key rollout (feature-flagged today) — ready-for-review ≠ merge-now.

dustinbyrne (approved):
there's still a bunch of references to the now deprecated `personal_api_key`/`POSTHOG_PERSONAL_API_KEY` in examples and generators

PR #213, opened 2026-07-02
feat: walk exception cause chains

hpouillot:

Capture the full Ruby `exception.cause` chain instead of only the outermost exception, preserving parent linkage and mechanism metadata for each entry. This also lets automatic Rails integrations set the mechanism type and handled flag for captured exceptions.

This is the first PR split out of the broader error-tracking capture improvements.

Not run successfully locally. `bundle exec rspec spec/posthog/exception_capture_spec.rb spec/posthog/client_spec.rb spec/posthog/rails/exception_mechanism_spec.rb` failed because the `rspec` executable is not installed in the local bundle.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi. The original combined error-tracking capture change was split so this PR only covers cause-chain walking and capture mechanism metadata; the path-stabilization change is left for a follow-up PR.

### DIFF
--- a/.changeset/eager-owls-chain.md
+++ b/.changeset/eager-owls-chain.md
@@ -1,0 +1,9 @@
+---
+"posthog-ruby": minor
+"posthog-rails": minor
+---
+
+Improve error tracking capture signals:
+
+- `capture_exception` now walks the full `exception.cause` chain (outermost-first, cycle-safe, capped at 50) instead of reporting only the outermost exception; chained causes are tagged with a `chained` mechanism and parent linkage.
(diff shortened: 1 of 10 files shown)


cat-ph (approved):
interesting, I thought we had this, thanks for adding!

looks good, maybe warrants a stamp from @PostHog/team-client-libraries too because of the public api change but I think it's additive only and shouldn't break existing users

turnipdabeets (approved):

ablaszkiewicz (approved):
🚀

PR #214, opened 2026-07-02
feat: use relative frame filenames

### LINKED ISSUE #212: Error tracking: stack trace `filename` is reduced to `File.basename`, dropping the project-relative path
### Version

`posthog-ruby` 3.11.0; `posthog-rails` 3.11.0; Ruby 3.3.5

### Steps to Reproduce

1. Configure error tracking per the Rails docs (`PostHog::Rails.configure { |c| c.auto_capture_exceptions = true }` + `PostHog.init`).
2. Raise any exception from an application file nested in a directory, e.g. `app/services/foo/bar.rb`.
3. Let the exception be auto-captured (or call `PostHog.capture_exception(e)` directly).
4. Inspect the resulting event in the PostHog Error Tracking UI stack trace view.

### Expected Result

Each stack frame's `filename` carries the project-relative path (e.g. `app/services/foo/bar.rb`), (matching the way sentry-ruby works), where `filename` is project-root-relative. This is what the Error Tracking UI renders in the frame header, so frames should be identifiable without collision.

### Actual Result

`filename` contains only the bare file name (`bar.rb`) because `parse_backtrace_line` sets it to `File.basename(file)`:

```ruby
# lib/posthog/exception_capture.rb
frame = {
  'filename' => File.basename(file),  # only the bare file name
  'abs_path' => file,                 # full path is captured here
  'lineno'   => lineno,
  'function' => method_name,
  'in_app'   => !gem_path?(file),
  'platform' => 'ruby'
}
```

The full path is still transmitted in `abs_path`, but the UI displays `filename`, so every frame shows only a basename. Files with the same basename in different directories are indistinguishable in the stack trace view.

Note: this file is derived from `sentry-ruby` (credited in the header). Sentry computes `filename` in `StacktraceInterface::Frame#compute_filename` (`sentry/interfaces/stacktrace.rb`) by stripping a leading prefix from `abs_path` — the project root for in-app frames, otherwise the longest matching `$LOAD_PATH` entry — so directory structure is always preserved (e.g. `app/services/foo/bar.rb`). It never reduces `filename` to a basename. The `File.basename` call therefore looks like an unintended divergence from the ported behavior.

hpouillot:

Use project-relative stack frame `filename` values while keeping the raw path in `abs_path`. Cymbal uses `filename` as the source path for Ruby raw frames, so this keeps fingerprints stable across deploy directories without dropping the original absolute path.

Stacked on #213.
Fixes: https://github.com/PostHog/posthog-ruby/issues/212

Not run successfully locally. `bundle exec rspec spec/posthog/exception_capture_spec.rb spec/posthog/client_spec.rb spec/posthog/rails/exception_mechanism_spec.rb` failed because the `rspec` executable is not installed in the local bundle.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi. This follow-up PR keeps the path-stabilization changes separate from the exception-chain work and aligns Ruby raw frame payloads with Cymbal's `filename`/`abs_path` expectations.


ablaszkiewicz (approved):

.changeset/eager-owls-chain.md
```diff
@@ -7,3 +7,4 @@ Improve error tracking capture signals:
 
 - `capture_exception` now walks the full `exception.cause` chain (outermost-first, cycle-safe, capped at 50) instead of reporting only the outermost exception; chained causes are tagged with a `chained` mechanism and parent linkage.
 - Exception mechanisms now reflect the capture source: manual captures stay `generic`/`handled: true`, while the Rails middleware, `Rails.error` subscriber, and ActiveJob integrations tag captures as `rails`/`rails_error_reporter`/`active_job` with the correct `handled` flag. `capture_exception` accepts a `mechanism:` keyword.
+- Stack frame `filename` values are now project-relative paths (relative to the Rails root or `Dir.pwd`, falling back to the basename outside the project) so fingerprints stay stable across per-host deploy directories; `abs_path` keeps the raw absolute path. `in_app` is now based on gem/stdlib install locations (`Gem.path`, loaded gem specs, `rubylibprefix`) instead of path substrings.
```
cat-ph: I think this will need a different changeset if the other one gets merged separately

cat-ph (approved):

.changeset/eager-owls-chain.md
turnipdabeets: agree on this, could we fix this before merging into #213?

turnipdabeets (approved):

PR #219, opened 2026-07-15
feat(flags): add $feature_flag_has_experiment to $feature_flag_called events

haacked:

Part of a cross-SDK effort to shrink `$feature_flag_called` events. This first phase adds a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, reflecting the server's `has_experiment` signal: `metadata.has_experiment` in the `/flags?v=2` response and `has_experiment` on `/api/feature_flag/local_evaluation` flag definitions. When the server does not report the field (older deployments, bootstrapped flags), the property is omitted, so it is tri-state: `true`, `false`, or absent (unknown).

This lets ingestion distinguish experiment-linked flag events (which need the full property set for exposure analysis) from the rest, and lets us measure the split before a later phase strips the expensive properties (`$feature/<key>`, `$feature_flag_payload`, per-event system metadata) from non-experiment flag events. This PR intentionally minimizes nothing: no properties are removed, no config options are added, and dedupe behavior is unchanged.

## Changes

- `FeatureFlagMetadata` parses `has_experiment` from `/flags?v=2` metadata (boolean-coerced, default `false`).
- Poller `get_feature_flag` returns `has_experiment` sourced from the stored `/flags/definitions` definition on local eval and from response metadata on remote fallback; `get_all_flags_and_payloads` carries parsed flag objects through an internal key (stripped from the public return) to reach metadata on the bulk remote path.
- `client.rb` adds the property on the single-flag path and threads it into `EvaluatedFlagRecord`s; snapshot-path events include it (missing flags omit the property).
- Public API snapshot regenerated; minor changeset.

`bundle exec rspec`: 614 examples, 0 failures, with new coverage for remote true/false/absent/missing-flag, local-definition true/absent, single-flag local and remote v4/v3 paths, and metadata parsing; five existing exact-hash assertions updated. `rubocop` clean.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/feature-flag-has-experiment.md
+++ b/.changeset/feature-flag-has-experiment.md
@@ -1,0 +1,5 @@
+---
+"posthog-ruby": minor
+---
+
+feat: add a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, sourced from the server-reported `has_experiment` field (defaults to false when the server does not report it)
--- a/.changeset/feature-flag-has-experiment.md
+++ b/.changeset/feature-flag-has-experiment.md
@@ -2,4 +2,4 @@
 "posthog-ruby": minor
 ---
 
-feat: add a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, sourced from the server-reported `has_experiment` field (defaults to false when the server does not report it)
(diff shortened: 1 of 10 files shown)


turnipdabeets (approved):
LGTM! Every read is nil-safe (`&.` chains / short-circuits), the internal `flagDetails` key is stripped before the public return, and true/false/absent are covered on the metadata, single-flag local/remote, and snapshot paths.

PR #222, opened 2026-07-17
fix(flags): pass the resolved polling interval to the definitions poll timer

flivni:

`feature_flags_polling_interval` is documented as "Defaults to 30" and the poller computes `@polling_interval = polling_interval || 30`, but the value was not actually used, and the 60-second concurrent-ruby was used instead.

This PR fixes the above.

Behavior change worth reviewing: Halves the current effective default. Alternatively, we could update the documented default to 60s.

- New "definitions polling interval" specs (2 examples) assert the TimerTask's effective `execution_interval`: 30 when the option is unset, and the configured value when provided. Before the fix, the default case reports 60.
- Full suite: 616 examples, 0 failures. `rubocop` clean. `rake public_api:check` passes with the snapshot regenerated for the one new constant.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Autonomy: Agent-written, human-directed and reviewed

This bug was found by Claude Code (Fable 5) in a code review on a different PR. All code was written by Claude. My contribution is limited to the "Motivation and Context" paragraph and this paragraph here. I've read the code and I fully understand it.

### DIFF
--- a/.changeset/polling-interval-default.md
+++ b/.changeset/polling-interval-default.md
@@ -1,0 +1,5 @@
+---
+"posthog-ruby": patch
+---
+
+fix: honor the documented 30s default for `feature_flags_polling_interval`. The resolved default was computed but never passed to the polling timer, so when the option was unset the effective interval was concurrent-ruby's 60s TimerTask default. Definitions now poll every 30s by default as documented; set `feature_flags_polling_interval: 60` to keep the previous effective cadence.
--- a/.changeset/polling-interval-default.md
+++ b/.changeset/polling-interval-default.md
@@ -2,4 +2,4 @@
 "posthog-ruby": patch
 ---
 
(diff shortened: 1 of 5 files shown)


.changeset/polling-interval-default.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-ruby": patch
+---
+
+fix: honor the documented 30s default for `feature_flags_polling_interval`. The resolved default was computed but never passed to the polling timer, so when the option was unset the effective interval was concurrent-ruby's 60s TimerTask default. Definitions now poll every 30s by default as documented; set `feature_flags_polling_interval: 60` to keep the previous effective cadence.
```
turnipdabeets: same here, This is the public facing message so we don’t need `fix`

flivni: Thanks. Done.

turnipdabeets (approved):

turnipdabeets:
LGTM but @flivni both PRs need a verified signature to merge. Mind adding those? We'll also need to rebase #221 after this one merges

flivni:
@turnipdabeets Done — signatures added on both PRs. Can rebase #221 once this one merges.

PR #230, opened 2026-08-03
feat: Add timeout: option to flush and shutdown

flivni:
`posthog_client.shutdown()` lets you cleanly shut down. This is particularly useful when sync mode is false (asynchronous mode) as it allows you to wait until the client has flushed its queue and stopped its worker threads. 

The problem is... the call can block for a long time or even indefinitely. In certain environments, e.g. shutting down an app server worker, a long block will cause timeout thresholds to be reached, resulting in a hard kill of the app server process, interrupting any other subsequent shutdown steps.

This PR adds an optional `timeout` argument to `shutdown(...)` allowing the user to specify a maximum time to spend flushing the worker queue before giving up and shutting down the worker thread and exiting cleanly. 

Note: `posthog-node` already supports same with `shutdown(shutdownTimeoutMs)`.

E.g.

```ruby
  on_worker_shutdown do
    posthog_client.shutdown(timeout: 4.0)

    # ... additional shutdown steps
  end
```

* if `timeout` isn't specified, the previous behavior is preserved.
* `shutdown(..)` and `flush(..)` both get the new optional `timeout` argument and they both now return `Boolean` rather than `void`. `false` means not shut down (or flushed) completely. 
* `shutdown(..)` uses the full timeout value for flushing the queue, then waits up to one additional second for worker and transport cleanup.
* when sync_mode is true, `timeout` has no effect and both methods always return `true`.

### Asynchronous mode only
The timeout value only affects `shutdown` and `flush` when sync_mode is false (asynchronous mode). This is the primary use case. But, one might desire the ability (and the symmetry) for timeout to be used when sync_mode is true as well. I didn't tackle this for a few reasons:
1. It has lower value / need than the async case
2. It's a bigger change than the async case
3. There are a few behavior changes that would likely need to be made. One specific gotcha is that `on_error` is called within the context of the `sync_lock` mutex and can in turn call `flush`, making it re-entrant.

- New specs in `client_spec.rb` (9 examples): async `flush(timeout:)` returns `true` when a real worker drains the queue (stubbed transport) and `false` with a stuck worker, leaving events queued; the same pair for `shutdown(timeout:)`; sync mode accepts and ignores the timeout on both methods, with `shutdown` still closing the transport. Shutdown coverage also verifies in-flight completion, concurrent callers, and cleanup of a timed-out worker. The success-path specs use a real `SendWorker` and worker thread so the deadline loop is actually exercised rather than short-circuited.
- Full suite: 647 examples, 0 failures. `rubocop` clean. `rake public_api:check` passes with the snapshot regenerated for the two changed signatures.
- End-to-end smoke test driving the library as an app would:

| Scenario | Call | Result |
|---|---|---|
| Unroutable endpoint (transport would retry ~2 min) | `shutdown(timeout: 2)` | `false` after the 2.0s flush timeout, plus up to 1.0s for cleanup |
| Unroutable endpoint | `flush(timeout: 1)` | `false` in 1.0s, client keeps running |
| Healthy (stubbed) transport, queued event | `shutdown(timeout: 5)` | `true` in 0.1s |
| Healthy, no-arg `flush` then `shutdown` (backwards compat) | `flush` / `shutdown` | `true` / `true` in 0.1s |

## :pencil: Checklist

- [X] I reviewed the submitted code.
- [X] I added tests to verify the changes.
- [X] I updated the docs if needed.
- [X] No breaking change or entry added to the changelog.

### If releasing new changes

- [X] Ran `pnpm changeset` to generate a changeset file

## 🤖 Agent context

Autonomy: Human-driven (Claude Fable 5 assisted)

Fully designed / structured by human. Methods defined and documented by human.

Implementation written by AI with full line-level human review. Tests fully written by AI with minimal human review.

PR description: Motivation & Context and Agent Context sections written by human.  The rest of the PR description written by AI.

The human understands all the code in this PR.


marandaneto:
thanks @flivni 
i pushed some changes, lets see if ci is happy

marandaneto:
@ioannisj or @dustinbyrne can you take a look since i did some changes

ioannisj:
> @ioannisj or @dustinbyrne can you take a look since i did some changes

Look good to my eyes but if not urgent maybe we'll need @dustinbyrne trained eye as well

marandaneto (approved):

Issue #187, opened 2026-06-18
Treat all 2xx ingestion responses as successful


marandaneto:
## Summary
Treat all 2xx ingestion responses as successful in the Ruby SDK.

## Spec
- `sdk-specs/openspec/specs/http-client/spec.md`

## Current evidence
The SDK audit found send paths treating only `status == 200` as success; the spec expects any 2xx response to be considered successful.

## Expected behavior
- Classify HTTP 2xx responses as success for ingestion sends.
- Avoid invoking error handling for successful 2xx responses.
- Preserve existing handling for retryable/non-retryable failures.
- Add tests for non-200 2xx statuses, e.g. 201/202/204.

marandaneto:
https://github.com/PostHog/posthog-ruby/pull/253

PR #253, opened 2026-08-19
fix: treat 2xx responses as successful

saidM:

Only `200` responses were considered successful, so other successful responses such as `201`, `202`, and `204` triggered `on_error`.

This updates both the sync client and background worker to accept any 2xx response.

It fixes issue https://github.com/PostHog/posthog-ruby/issues/187

Added specs covering `201`, `202`, and `204` responses in both code paths.

- Full test suite: 670 examples, 0 failures
- RuboCop: no offenses
- `rake public_api:check`: passed

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Used GPT-5.6 in Codex to help implement and test the change. I reviewed the final code.

### DIFF
--- a/spec/posthog/client_spec.rb
+++ b/spec/posthog/client_spec.rb
@@ -273,6 +273,20 @@ module PostHog
         expect(error_status).to eq(400)
       end
 
+      [201, 202, 204].each do |status|
+        it "does not call on_error for a #{status} response" do
+          on_error = proc {}
+          expect(on_error).to_not receive(:call)
+
+          allow_any_instance_of(PostHog::Transport).to(
+            receive(:send).and_return(PostHog::Response.new(status))
+          )
+
+          sync_client = Client.new(api_key: API_KEY, sync_mode: true, on_error: on_error)
+          sync_client.capture(Queued::CAPTURE)
+        end
+      end
+
       it 'flush does not attempt to run a worker' do
         sync_client = Client.new(api_key: API_KEY, sync_mode: true)
         expect(sync_client).not_to receive(:ensure_worker_running)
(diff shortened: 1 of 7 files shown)


spec/posthog/client_spec.rb
```diff
@@ -273,6 +273,20 @@ def shutdown
...
       end
 
+      [201, 202, 204].each do |status|
+        it "does not call on_error for a #{status} response" do
+          on_error = proc {}
+          expect(on_error).to_not receive(:call)
+
+          allow_any_instance_of(PostHog::Transport).to(
+            receive(:send).and_return(PostHog::Response.new(status))
+          )
```
dustinbyrne, line 283: a real `Net::HTTPNoContent` (204) response has a `nil` body and we JSON parse the body in the transport (lib/posthog/transport.rb), so in this case, it'd actually raise an exception

we should fix / test this

dustinbyrne (commented):
thanks for putting this together @saidM! one small bug described below:

dustinbyrne:
one more thing: check RELEASING.md for release prerequisites. [verified commits](https://docs.github.com/en/authentication/managing-commit-signature-verification) are also required for us to be able to merge this change.

spec/posthog/client_spec.rb
saidM: hey @dustinbyrne, good catch!  going to push a fix shortly

dustinbyrne:
code looks good here, thanks for the update! just waiting on signed commit

saidM:
great, done just now!

dustinbyrne (approved):
thanks again!

PR #256, opened 2026-08-20
fix: Drop events when before_send callbacks raise

marandaneto:

A `before_send` callback that raised an exception caused the SDK to continue with the original event. This could send data that the callback intended to redact or reject.

This change logs the callback error and drops the event instead. Both queued and synchronous capture paths return without enqueueing or sending the event, including when the callback mutates it before raising. The Rails log callback documentation is updated to match, and the changeset includes patch releases for both `posthog-ruby` and `posthog-rails`.

- `bundle exec rspec spec/posthog/client_spec.rb` - 127 examples passed
- `bundle exec rubocop lib/posthog/client.rb posthog-rails/lib/posthog/rails/logs/appender.rb spec/posthog/client_spec.rb`
- `bundle exec rake public_api:check`
- `pnpm changeset status`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent helped audit the existing behavior, update the implementation and tests, run validation, and prepare this PR. The chosen behavior follows the cross-SDK contract: callback failures are contained and the event is dropped rather than sent without the callback's intended processing.

### DIFF
--- a/.changeset/quiet-hooks-close.md
+++ b/.changeset/quiet-hooks-close.md
@@ -1,0 +1,5 @@
+---
+'posthog-ruby': patch
+---
+
+Drop events when before_send callbacks raise exceptions.
--- a/.changeset/quiet-hooks-close.md
+++ b/.changeset/quiet-hooks-close.md
@@ -1,5 +1,6 @@
 ---
 'posthog-ruby': patch
+'posthog-rails': patch
 ---
 
 Drop events when before_send callbacks raise exceptions.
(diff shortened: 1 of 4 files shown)


turnipdabeets (approved):
Approving. `enqueue` already guards `return false if action.nil? || action.empty?` before the `sync_mode` branch, so the queued and synchronous paths both drop correctly, and both are covered. Good call folding in the Rails appender comment — it was the only place still documenting the old fail-open behavior, so it would have gone stale silently.

Issue #189, opened 2026-06-18
Reset feature-flag-called tracker on flag reload


marandaneto:
## Summary
Reset feature-flag-called dedupe tracker when flags are reloaded in the Ruby SDK.

## Spec
- `sdk-specs/openspec/specs/feature-flag-called-tracker/spec.md`

## Current evidence
The SDK audit found the dedupe map is not cleared by `reload_feature_flags` / flag definition reload.

## Expected behavior
- Clear/reset `$feature_flag_called` dedupe state when feature flags/definitions are reloaded or reset.
- Ensure the next access after a reload can emit fresh tracking events.
- Preserve duplicate suppression for unchanged repeated reads between reloads.
- Add tests covering tracking before and after reload.

PR #267, opened 2026-09-05
fix: reset flag-called tracking when definitions reload

marandaneto:

Related to #189.

After local flag definitions reload, the client still remembers earlier `$feature_flag_called` events and suppresses the next access. Clear the existing tracker when definitions are applied or when a quota response discards loaded definitions. Manual reloads, background polling, and external-cache refreshes use the same callback.

Definition publication and tracker reset now share the existing deduplication mutex, closing the race where a refresh erased an event recorded against newly published definitions. Network requests, cache-provider callbacks, and logging stay outside that lock. Repeated reads between reloads remain deduplicated. HTTP 304 responses, failed requests, and responses without definitions leave tracking unchanged. Evaluations already in flight may still complete using definitions they read before a refresh.

- Added 13 regression examples that inspect queued events, covering manual and background reloads, external-cache refreshes, empty definitions, quota discard, shared snapshot tracking, failure preservation, and concurrent duplicate reads.
- Reproduction against the unchanged base failed 8 of those 13 examples. All 13 pass with this fix. Ten seeded regression runs also passed during implementation.
- Added six deterministic concurrency cases spanning network refreshes, external-cache refreshes, and quota clearing through both single-flag and snapshot access. All six reproduced duplicate events before the follow-up fix (expected one event, got two), and pass afterward.
- Re-ran `bundle exec rspec`: 696 examples, 0 failures. The 19-example reload suite also passed ten seeded runs (190 examples).
- Re-ran `bundle exec rubocop` (85 files), `bundle exec rake public_api:check`, and the core gem build. All passed. The gem build retains its existing missing-author warning.
- Original branch autoreview passed for `4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96`. Local autoreview of the concurrency follow-up, published as `d0b0315e801c64c409e38c3555619950e8f6eb74`, also passed with no actionable findings.

Validation used Ruby 4.0.6 and Bundler 4.0.13 with frozen dependencies. The CI Ruby 3.2/3.3/3.4 matrix was not run locally. Tests use WebMock rather than a live PostHog service.

## :pencil: Checklist

- [x] I updated the docs if needed. No documentation change was needed.

### If releasing new changes

The changeset selects patch releases for both `posthog-ruby` and `posthog-rails`. Rails delegates feature-flag access to the core client and pins the exact core SDK version, so it needs a release to receive this fix. The changeset was edited directly; the generator was not run.

## 🤖 Agent context

Pi agents implemented and independently reviewed this user-directed fix. Tools used were Git, Bundler/RSpec, RuboCop, Rake, RubyGems, GitHub CLI, and the isolated Pi autoreview helper. Session references: implementation `3efed453-bba9-4ebc-853b-b7247693a8d8`, publication `510ebebc-a750-45a5-88ce-92821ccd3256` (local sessions, no public link).

The fix keeps the existing bounded tracker and mutex rather than adding generation tracking or changing the public client API. Human review is required before merging.

### DIFF
--- a/.changeset/fresh-flag-reload-tracking.md
+++ b/.changeset/fresh-flag-reload-tracking.md
@@ -1,5 +1,6 @@
 ---
 "posthog-ruby": patch
+"posthog-rails": patch
 ---
 
 Reset feature flag event deduplication when local flag definitions are refreshed or discarded, allowing the next flag access to emit a fresh event.
--- a/.changeset/fresh-flag-reload-tracking.md
+++ b/.changeset/fresh-flag-reload-tracking.md
@@ -1,0 +1,5 @@
+---
+"posthog-ruby": patch
+---
+
+Reset feature flag event deduplication when local flag definitions are refreshed or discarded, allowing the next flag access to emit a fresh event.
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -210,11 +210,8 @@ module PostHog
             feature_flag_request_max_retries: opts[:feature_flag_request_max_retries],
             async_load: opts[:feature_flags_async_load] == true,
             user_agent: @headers['User-Agent'],
-            on_flag_definitions_updated: lambda {
-              @distinct_id_has_sent_flag_calls_mutex.synchronize do
-                @distinct_id_has_sent_flag_calls.clear
-              end
-            }
+            flag_definitions_update_mutex: @distinct_id_has_sent_flag_calls_mutex,
+            on_flag_definitions_updated: -> { @distinct_id_has_sent_flag_calls.clear }
           )
       end
 
--- a/lib/posthog/client.rb
+++ b/lib/posthog/client.rb
@@ -191,6 +191,12 @@ module PostHog
         end
       end
 
+      # Initialize tracking before the poller can load definitions (including async loads).
+      @distinct_id_has_sent_flag_calls_mutex = Mutex.new
+      @distinct_id_has_sent_flag_calls = SizeLimitedHash.new(Defaults::MAX_HASH_SIZE) do |hash, key|
+        hash[key] = SizeLimitedArray.new(Defaults::MAX_HASH_SIZE)
+      end
+
       unless @disabled
         @feature_flags_poller =
           FeatureFlagsPoller.new(
@@ -203,13 +209,13 @@ module PostHog
             flag_definition_cache_provider: opts[:flag_definition_cache_provider],
             feature_flag_request_max_retries: opts[:feature_flag_request_max_retries],
             async_load: opts[:feature_flags_async_load] == true,
-            user_agent: @headers['User-Agent']
+            user_agent: @headers['User-Agent'],
+            on_flag_definitions_updated: lambda {
+              @distinct_id_has_sent_flag_calls_mutex.synchronize do
+                @distinct_id_has_sent_flag_calls.clear
+              end
+            }
           )
-      end
-
-      @distinct_id_has_sent_flag_calls_mutex = Mutex.new
-      @distinct_id_has_sent_flag_calls = SizeLimitedHash.new(Defaults::MAX_HASH_SIZE) do |hash, key|
-        hash[key] = SizeLimitedArray.new(Defaults::MAX_HASH_SIZE)
       end
 
       @before_send = opts[:before_send]
--- a/lib/posthog/feature_flags.rb
+++ b/lib/posthog/feature_flags.rb
@@ -45,7 +45,9 @@ module PostHog
     #   immediate first tick at construction, then the regular polling cadence, which keeps retrying until a
     #   load succeeds.
     # @param user_agent [String] User-Agent header sent with feature flag requests.
-    # @param on_flag_definitions_updated [Proc, nil] Internal callback after definitions are applied or discarded.
+    # @param flag_definitions_update_mutex [Mutex] Internal lock shared with flag-called deduplication.
+    # @param on_flag_definitions_updated [Proc, nil] Internal callback after definitions are applied or discarded,
+    #   called while holding flag_definitions_update_mutex.
     def initialize(
       polling_interval,
       secret_key,
@@ -57,7 +59,8 @@ module PostHog
       feature_flag_request_max_retries: nil,
       async_load: false,
       user_agent: "posthog-ruby/#{PostHog::VERSION}",
-      on_flag_definitions_updated: nil
+      on_flag_definitions_updated: nil,
+      flag_definitions_update_mutex: Mutex.new
     )
       @polling_interval = polling_interval || Defaults::FeatureFlags::POLLING_INTERVAL_SECONDS
       @secret_key = secret_key
@@ -78,6 +81,7 @@ module PostHog
       @async_load = async_load
       @user_agent = user_agent
       @on_flag_definitions_updated = on_flag_definitions_updated
+      @flag_definitions_update_mutex = flag_definitions_update_mutex
       # Server-controlled gate for minimal `$feature_flag_called` events, read
       # from the top-level `minimal_flag_called_events` key of the local
       # evaluation definitions payload. false when the server does not send it.
@@ -1220,20 +1224,22 @@ module PostHog
 
       # Handle quota limits with 402 status
       if res.is_a?(Hash) && res[:status] == 402
-        definitions_were_loaded = definitions_loaded?
         logger.warn(
           '[FEATURE FLAGS] Feature flags quota limit exceeded - unsetting all local flags. ' \
           'Learn more about billing limits at https://posthog.com/docs/billing/limits-alerts'
         )
-        @feature_flags = Concurrent::Array.new
-        @feature_flags_by_key = {}
-        @group_type_mapping = Concurrent::Hash.new
-        @cohorts = Concurrent::Hash.new
-        @flag_definitions_loaded_at.value = nil
-        @minimal_flag_called_events = false
-        @loaded_flags_successfully_once.make_false
-        @quota_limited.make_true
-        @on_flag_definitions_updated&.call if definitions_were_loaded
+        @flag_definitions_update_mutex.synchronize do
+          definitions_were_loaded = definitions_loaded?
+          @feature_flags = Concurrent::Array.new
+          @feature_flags_by_key = {}
+          @group_type_mapping = Concurrent::Hash.new
+          @cohorts = Concurrent::Hash.new
+          @flag_definitions_loaded_at.value = nil
+          @minimal_flag_called_events = false
+          @loaded_flags_successfully_once.make_false
+          @quota_limited.make_true
+          @on_flag_definitions_updated&.call if definitions_were_loaded
+        end
         return
       end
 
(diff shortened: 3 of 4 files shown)


marandaneto:
maybe a better approach would be to check the results and the cached flags and only clear the values that have changed, wdyt? or is it ok to emit again? i recall those events being "expensive" but later this was improved during ingestion

lib/posthog/feature_flags.rb
```diff
@@ -1280,6 +1285,7 @@ def _apply_flag_definitions(data)
       logger.debug "Loaded #{@feature_flags.length} feature flags and #{@cohorts.length} cohorts"
       @flag_definitions_loaded_at.value = (Time.now.to_f * 1000).to_i
       @loaded_flags_successfully_once.make_true if @loaded_flags_successfully_once.false?
+      @on_flag_definitions_updated&.call
```
haacked: `blocking`: The `$feature_flag_called` dedup tracker gets wiped on every successful application of flag definitions, even when those definitions are identical to what's already loaded. `_apply_flag_definitions` fires the reset callback at lib/posthog/feature_flags.rb:1288 with no comparison against the previously loaded values. That hurts most on a deployment configured with `flag_definition_cache_provider` (default poll interval 30 seconds), where `should_fetch_flag_definitions?` returns false on most ticks for every worker except the one holding the distributed fetch lock, so those followers reapply the same unchanged cached definitions tick after tick.

The external-cache branch of `_load_feature_flags` ([lines 1185-1191](https://github.com/posthog/posthog-ruby/blob/4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96/lib/posthog/feature_flags.rb#L1185-L1191)) calls `_apply_flag_definitions(cached_data)` whenever the provider says not to fetch, with none of the ETag or diff checking the direct-HTTP path has. `_apply_flag_definitions` then reassigns `@feature_flags`, `@feature_flags_by_key`, `@group_type_mapping`, and `@cohorts`, and reaches line 1288, where it calls `@on_flag_definitions_updated&.call` regardless of whether any of those new values differ from the old ones. That callback is [the lambda that clears the tracker](https://github.com/posthog/posthog-ruby/blob/4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96/lib/posthog/client.rb#L213-L217), which takes `@distinct_id_has_sent_flag_calls_mutex` and calls `.clear` on `@distinct_id_has_sent_flag_calls`, emptying the whole dedup tracker. The poller's timer runs `_load_feature_flags` again every 30 seconds by default, so every non-leader worker in a multi-worker deployment empties its tracker on that cadence.

The unconditional fire also hits the very first successful load. Under `feature_flags_async_load: true` the timer performs the initial load in the background, so reads that arrive before it completes fall back to remote evaluation and record their `$feature_flag_called` keys in the tracker; when that first load then succeeds, `_apply_flag_definitions` fires the callback and clears exactly those entries, so a subsequent identical read emits a duplicate. The same sequence plays out when early polls fail and a later one succeeds.

The net effect is that `$feature_flag_called` re-emits for every active distinct_id/flag/value combination roughly every 30 seconds on cache-provider follower workers, instead of once per process lifetime, multiplying billable event volume for exactly the deployment shape the cache provider exists to serve. The direct-HTTP-fetch path doesn't have this problem: the server sends an ETag and a 304 response short-circuits before `_apply_flag_definitions` ever runs.

Compare the newly applied definitions (flags by key, `group_type_mapping`, `cohorts`) against what's already loaded, and fire the callback only when they differ. Also gate on whether definitions were already loaded before this apply, using the same `definitions_loaded?` check the 402 branch already makes at [line 1223](https://github.com/posthog/posthog-ruby/blob/4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96/lib/posthog/feature_flags.rb#L1223), so the very first successful load doesn't itself trigger a needless reset that can invalidate already-emitted events.

For coverage, add a test where the external-cache provider returns identical definitions across two ticks and asserts no new `$feature_flag_called` event is emitted. That's the inverse of [the existing spec](https://github.com/posthog/posthog-ruby/blob/4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96/spec/posthog/feature_flag_called_reload_spec.rb#L101-L119), which currently feeds back the same definitions and asserts a reset happens anyway.

def _apply_flag_definitions(data)
  flags = get_by_symbol_or_string_key(data, 'flags') || []
  group_type_mapping = get_by_symbol_or_string_key(data, 'group_type_mapping') || {}
  cohorts = get_by_symbol_or_string_key(data, 'cohorts') || {}
  minimal_flag_called_events = get_by_symbol_or_string_key(data, 'minimal_flag_called_events')

  new_flags = Concurrent::Array.new(flags.map { |f| deep_symbolize_keys(f) })
  new_by_key = {}
  new_flags.each { |flag| new_by_key[flag[:key]] = flag unless flag[:key].nil? }
  new_group_type_mapping = Concurrent::Hash[deep_symbolize_keys(group_type_mapping)]
  new_cohorts = Concurrent::Hash[deep_symbolize_keys(cohorts)]

  definitions_changed =
    definitions_loaded? &&
    (new_by_key != @feature_flags_by_key ||
     new_cohorts != @cohorts ||
     new_group_type_mapping != @group_type_mapping)

  @feature_flags = new_flags
  @feature_flags_by_key = new_by_key
  @group_type_mapping = new_group_type_mapping
  @cohorts = new_cohorts
  @minimal_flag_called_events = minimal_flag_called_events == true

  logger.debug "Loaded #{@feature_flags.length} feature flags and #{@cohorts.length} cohorts"
  @flag_definitions_loaded_at.value = (Time.now.to_f * 1000).to_i
  @loaded_flags_successfully_once.make_true if @loaded_flags_successfully_once.false?
  @on_flag_definitions_updated&.call if definitions_changed
end

spec/posthog/feature_flag_called_reload_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+    it 'resets tracking when a quota-limited response discards definitions' do
+      2.times { client.get_feature_flag('missing', 'user', only_evaluate_locally: true) }
+      expect_single_event
+      stub_request(:get, definitions_endpoint).to_return(status: 402, body: '{}')
+
+      client.reload_feature_flags
+
+      expect(client.feature_flags_loaded?).to be(false)
+      2.times { client.get_feature_flag('missing', 'user', only_evaluate_locally: true) }
+      expect_single_event
```
haacked: The new spec covers a single 402 quota-limit response but never exercises a second consecutive 402, which is the only case the `definitions_were_loaded` guard on the 402 branch protects.

Extend the existing quota-limit example (spec/posthog/feature_flag_called_reload_spec.rb:132-142) to reload a second time while still quota-limited, and assert no further event is emitted:

```ruby
expect(client.feature_flags_loaded?).to be(false)
2.times { client.get_feature_flag('missing', 'user', only_evaluate_locally: true) }
expect_single_event

client.reload_feature_flags

2.times { client.get_feature_flag('missing', 'user', only_evaluate_locally: true) }
expect(client.queued_messages).to eq(0)
```

.changeset/fresh-flag-reload-tracking.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-ruby": patch
+---
+
+Reset feature flag event deduplication when local flag definitions are refreshed or discarded, allowing the next flag access to emit a fresh event.
```
haacked: The changeset describes the reset as a one-time effect on the next flag access, but doesn't mention that `$feature_flag_called` event volume increases on every reload cycle, which matters most for `flag_definition_cache_provider` deployments that reload every 30 seconds.

Reset `$feature_flag_called` deduplication when local flag definitions are applied or discarded, so the next access to each flag emits a fresh event. Deployments that reload definitions frequently, including those using `flag_definition_cache_provider`, will see more `$feature_flag_called` events than before; repeated reads between reloads are still deduplicated.

lib/posthog/client.rb
```diff
@@ -203,15 +209,15 @@ def initialize(opts = {})
             flag_definition_cache_provider: opts[:flag_definition_cache_provider],
             feature_flag_request_max_retries: opts[:feature_flag_request_max_retries],
             async_load: opts[:feature_flags_async_load] == true,
-            user_agent: @headers['User-Agent']
+            user_agent: @headers['User-Agent'],
+            on_flag_definitions_updated: lambda {
```
haacked: The three-line tracker-reset block (take the mutex, clear the tracker) is now duplicated byte-for-byte between the new `on_flag_definitions_updated` lambda and the pre-existing shutdown path.

private

def _reset_flag_called_tracking
  @distinct_id_has_sent_flag_calls_mutex.synchronize do
    @distinct_id_has_sent_flag_calls.clear
  end
end

# at the FeatureFlagsPoller.new call site:
on_flag_definitions_updated: method(:_reset_flag_called_tracking)

# at shutdown:
_reset_flag_called_tracking

haacked: There's an existing review comment on this PR (from @greptile-apps[bot], at [the same line](https://github.com/posthog/posthog-ruby/blob/4f4b1de269fc651c7a79a1ba0f09d4ed0c90bf96/lib/posthog/feature_flags.rb#L1288)) about an ordering race: new definitions become visible before the tracker clears, so a concurrent evaluation's dedup record can be erased right after its event was sent, causing one duplicate emission. That comment's suggested fix (tag every tracker entry with a definitions generation) is heavier than the problem needs.

Give the poller a `Concurrent::AtomicFixnum` generation counter that increments only when applied definitions actually change (reusing the diffing from the other blocking finding on this line). Expose it with a reader next to `flag_definitions_loaded_at`. Then, inside the same mutex `_capture_feature_flag_called_if_needed` already holds, compare the client's last-seen generation to the poller's current one and clear the tracker there if they differ, before checking or marking the dedup key:

```ruby
should_capture = @distinct_id_has_sent_flag_calls_mutex.synchronize do
  generation = @feature_flags_poller&.flag_definitions_generation
  if generation != @flag_calls_definitions_generation
    @flag_calls_definitions_generation = generation
    @distinct_id_has_sent_flag_calls.clear
  end
  sent_keys = @distinct_id_has_sent_flag_calls[distinct_id]
  # ... existing dedup check
end
```

This closes the race because the clear and the dedup check/mark now happen in the same critical section on the reading thread, so no reset can wipe a record whose event already went out. It also removes the need for the `on_flag_definitions_updated` callback, both call sites, and the `definitions_loaded?` / `definitions_were_loaded` guards entirely.

lib/posthog/feature_flags.rb
```diff
@@ -45,6 +45,7 @@ class FeatureFlagsPoller
     #   immediate first tick at construction, then the regular polling cadence, which keeps retrying until a
     #   load succeeds.
     # @param user_agent [String] User-Agent header sent with feature flag requests.
+    # @param on_flag_definitions_updated [Proc, nil] Internal callback after definitions are applied or discarded.
```
haacked: The new doc comment says the callback fires "after definitions are applied or discarded," but a repeated 402 while already quota-limited discards definitions again without firing the callback, since the discard-fire is itself gated on `definitions_loaded?`.

Callback invoked after definitions are applied, and after a quota-limited response discards previously loaded definitions.

dustinbyrne:
> maybe a better approach would be to check the results and the cached flags and only clear the values that have changed, wdyt? or is it ok to emit again? i recall those events being "expensive" but later this was improved during ingestion

~we could also key off distinct id _and_ value - so if the evaluation result changes it'll be emitted, otherwise it's deduplicated immediately. i think we're doing this elsewhere, i have an agent looking into it.~

we're already doing this. ingestion does deduplicate these as well.

dustinbyrne (approved):
The existing dedupe key already includes distinct ID, flag key, response and normalized groups. Resetting the tracker when definitions are applied matches its lifecycle contract; 304 and failed refreshes preserve tracking.

lib/posthog/feature_flags.rb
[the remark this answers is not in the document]
marandaneto: Fixed in d0b0315e801c64c409e38c3555619950e8f6eb74. Reproduced with deterministic queue barriers that pause the refresh before it acquires the deduplication lock, then evaluate concurrently and read again after refresh. All six cases (network, external cache, and quota clearing, through single-flag and snapshot access) emitted two events instead of one before the fix.

Definition publication/clearing and tracker reset now run under the same existing deduplication mutex, so an access to newly published definitions cannot be recorded and then erased by that reset. Network requests, cache-provider callbacks, and logging remain outside the lock. The existing bounded tracker and deduplication key/policy are unchanged.

`bundle exec rspec spec/posthog/feature_flag_called_reload_spec.rb`: 19 examples pass, including all six regressions; ten seeded runs pass (190 examples). Full suite: 696 examples pass. RuboCop, public API check, gem build, autoreview, and all 21 pushed-head CI checks pass, including Ruby 3.2/3.3/3.4.

PR #268, opened 2026-09-07
ci: cover native Ruby UUIDs and synchronous capture in SDK harness

dustinbyrne:

Exercise Ruby's own UUID generation and synchronous capture transport in the SDK harness, and include the UTC timestamp case from harness 1.0.0 in recurring CI.

- Observe SDK-generated UUIDs through the existing `before_send` hook; preserve explicitly supplied UUIDs.
- Run separate async and `sync_mode` profiles with distinct reports. Keep real snapshot flag evaluation and the native omitted GeoIP default.
- Keep assertion failures advisory while requiring both complete 47-case reports.

This changes only the adapter, its tests/documentation and CI. SDK public APIs/defaults and Rails are unchanged; no package release is needed.

- `bundle exec rspec`: 696 examples, 0 failures, including 19 adapter examples.
- `bundle exec rubocop`: 85 files, no offenses.
- `bundle exec rake public_api:check`: passed.
- Harness 1.0.0 against a locally built Ruby gem on Ruby 3.4.7:

| Profile | Selected | Passed | Failed |
| --- | ---: | ---: | ---: |
| async | 47 | 46 | 1 |
| sync | 47 | 45 | 2 |

Both profiles expose `feature_flags.request_payload.disable_geoip_omitted_defaults_to_false`: the SDK omits the field instead of sending literal false. Sync additionally fails `capture.batch_format.multiple_events_batched_together` because it sends each event immediately. Both profiles pass the UTC and native UUID definitions.

The full suites run without filtering. Sync's batch-retry identity definition passes on a retried single-event request; neither that result nor its multi-event batching failure certifies queue batching. V1, dedicated AI and non-gzip codecs remain unsupported. Existing diagnostic-state limitations are documented in the adapter README.

The inventory check accepts the real reports with these failures and rejects missing/zero-test reports. Docker was unavailable locally; container validation will come from this PR's CI.

## :pencil: Checklist

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted).

Implemented using a Pi coding agent with shell/file tools and GitHub CLI, followed by a fresh read-only agent review. Coverage uses existing public SDK calls and passive observation; wire assertions remain owned by the unchanged harness. Human review is still required.


dustinbyrne:
## Compliance failures and follow-up plan

At `d7b786af3c9c995189bf7b0bc46ed648ae00ce43`, the published harness 1.0.0 reports show **[async 46/47](https://github.com/PostHog/posthog-ruby/pull/268#issuecomment-5575265754)** and **[sync 45/47](https://github.com/PostHog/posthog-ruby/pull/268#issuecomment-5575264087)**.

| Failure | Classification and cause | Plan |
| --- | --- | --- |
| `feature_flags.request_payload.disable_geoip_omitted_defaults_to_false` — both profiles | **Harness issue.** The SDK omits `geoip_disable`; the assertion requires literal false even though the omitted value has the established false semantics. | Correct the shared optional-field assertion to accept that omission, preserving separate explicit-value tests and rejection of null/wrong values. |
| `capture.batch_format.multiple_events_batched_together` — sync only, five requests rather than one | **Harness applicability issue.** Synchronous capture sends each event inline; it does not accumulate a later queue batch. | Make shared batching tests execution-mode aware. Keep multi-event queue assertions on async and add appropriate singleton/retry checks for sync. Do not change synchronous SDK behavior merely to satisfy a queued-mode test. |

The sync batch-retry identity test's nominal pass currently inspects a retried singleton, so it is not proof of multi-event batching. The harness follow-up should also require real batch membership before certifying that behavior.

After those shared changes, rerun both complete profiles and reconcile their selected IDs. These are not outstanding adapter binding failures or a request to change Ruby defaults; the recorded failures stay visible while this adapter/CI PR is reviewed. [Profile boundaries](https://github.com/PostHog/posthog-ruby/blob/d7b786af3c9c995189bf7b0bc46ed648ae00ce43/sdk_compliance_adapter/README.md). Compliance assertions remain advisory despite green workflow checks.

turnipdabeets (approved):

PR #269, opened 2026-09-09
feat: experimental MCP analytics (PostHog::MCP) for the Ruby mcp gem

rafaeelaudibert:

PostHog ships MCP analytics for Node ([`@posthog/mcp`](https://github.com/PostHog/posthog-js/tree/main/packages/mcp)) and Python (`posthog.mcp`), but not Ruby. This adds `PostHog::MCP`, an **experimental** integration that wraps a server built on the official Ruby [`mcp`](https://rubygems.org/gems/mcp) gem so every tool call, handshake, listing, prompt, resource read, and failure lands in PostHog as a `$mcp_*` event with the same wire contract as the other two SDKs.

Setup is one line, and Rails apps that already call `PostHog.init` can omit the client:

```ruby
require 'posthog/mcp'
PostHog::MCP.instrument(server, posthog)   # or PostHog::MCP.instrument(server) with posthog-rails
```

What's included:

- `PostHog::MCP.instrument` with feature parity: injected `context` intent argument, conversation ids (prompt-back + `structuredContent` mirror), `get_more_tools`, `llm_model` capture, stateless `Mcp-Session-Id` tokens, transport identity headers, `identify`/`$identify`, `before_send`, `event_properties`, sanitization, PII redaction of intent, and 100KB truncation. Ruby also emits the prompt/resource events the other SDKs reserve.
- `PostHog::MCP::Client` for custom dispatchers (no `mcp` gem needed) and `PostHog::MCP::RackMiddleware` for custom Rack stacks.
- A private per-event `_lib`/`_lib_version` override in `Client#capture`. MCP events report `$lib: posthog-ruby-mcp`; unlike Node/Python this does **not** relabel the host client, so a Rails app's other events keep `posthog-rails`.
- `mcp` added to the Gemfile test group only (peer dependency, gemspec unchanged); public API snapshot now tracks `PostHog::MCP`.

Design notes for reviewers:

- The gem's documented `around_request` hook lacks the request, session, and params, so the integration prepends `MCP::Server#handle_request` and wraps the dispatch lambda it returns. That lambda's `params` is the same Hash the tool receives, which is what lets us strip injected arguments before the `tool.call(**args)` splat (an unknown keyword would otherwise become an opaque `-32603`). No `Thread.current`; the only ambient state is `Fiber[]` storage for the Streamable HTTP transport → server hop, because the gem re-parses the body in between.
- Experimental: a warning is logged on `require 'posthog/mcp'` and on `instrument` (stderr outside Rails, never stdout, since stdio servers own it). README, YARD, and the changeset say so too.
- Docs PR: PostHog/posthog.com (linked in a comment once open).

- `bundle exec rspec`: 812 examples, 0 failures (115 new examples under `spec/posthog/mcp`, including end-to-end specs against a real `MCP::Server` over `Server#handle` and over `StreamableHTTPTransport#call` in stateless and stateful modes, a concurrency spec, and the frozen cross-SDK vectors for FNV-1a session ids and the `Mcp-Session-Id` token copied from the Python/JS suites).
- `bundle exec rubocop`: clean. `bundle exec rake public_api:check`: passes.
- `examples/mcp_server.rb` driven over stdio with real `initialize` / `tools/list` / `tools/call` JSON-RPC lines; captured events dumped to stderr.
- Not done: sending events to a live PostHog project.

## :pencil: Checklist

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (`.changeset/quiet-ruby-mcp-analytics.md`, minor)

## 🤖 Agent context

Written with Claude Code (Claude Fable 5.1) from a plan the DRI reviewed and approved. The agent explored the JS and Python MCP packages and the Ruby `mcp` gem source, and the DRI made the calls on scope (full parity), wiring (explicit `instrument` only, no Railtie auto-wiring), and `$lib` attribution (per-event override rather than relabeling the client). The DRI also asked for the experimental marking and for avoiding `Thread.current`; the hook design was reworked from an `around_request` + thread-local draft to the single `handle_request` prepend described above. Cross-SDK divergences were resolved deliberately: Python's empty-`tools/list` error text and code-point FNV iteration, JS's `[Array]` truncation marker and requested-version era check for token minting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rafaeelaudibert:
Docs PR: https://github.com/PostHog/posthog.com/pull/20022

lib/posthog/mcp/truncation.rb
```diff
@@ -0,0 +1,298 @@
...
+    #    largest strings until under MAX_EVENT_BYTES.
+    #
+    # Pure functions; the input event is never mutated. Hash keys are strings.
+    #
+    # @api private
+    module Truncation
+      MAX_DEPTH = 10
+      MAX_BREADTH = 100
+      MAX_STRING_LENGTH = 32_768
+      MAX_EVENT_BYTES = 102_400
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed in b72f5e0. Valid: the core client drops messages over `Defaults::Message::MAX_BYTES` (32KB). Budget is now that minus 2KB headroom, and the size pass trims the largest strings before reducing depth. Regression spec asserts every queued message stays under the limit.

lib/posthog/mcp/identity.rb
```diff
@@ -0,0 +1,161 @@
...
+        return nil if identity.nil?
+
+        @cache[session_id] = identity
+      end
+
+      def set(session_id, identity)
+        @cache.delete(session_id)
+        @cache.shift if @cache.length >= @max_size && !@cache.key?(session_id)
+        @cache[session_id] = identity
+      end
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed in b72f5e0. `IdentityCache` now has its own mutex and `merge!` does the read-merge-write atomically; added a 16-thread concurrent identify spec.

lib/posthog/mcp/rack_middleware.rb
```diff
@@ -0,0 +1,102 @@
...
+            Log.debug(nil, "PostHog MCP session middleware: mint failed - #{e.message}")
+            nil
+          end
+          env[ENV_KEY] = SessionToken.decode(token) if token
+        end
+
+        status, headers, body = @app.call(env)
+        if token && headers.respond_to?(:key?) && headers.keys.none? { |k| k.to_s.casecmp?(MCP_SESSION_HEADER) }
+          headers[MCP_SESSION_HEADER] = token
+        end
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed in b72f5e0. The middleware attaches the token only on a 2xx response and clears `env['posthog_mcp.session']` otherwise; spec added.

lib/posthog/mcp/sink.rb
```diff
@@ -0,0 +1,86 @@
...
+        processed.each { |payload| dispatch(payload) }
+        processed
+      rescue StandardError => e
+        Log.debug(options, "Failed to capture PostHog event: #{e.message}")
+        []
+      end
+
+      # Runs the full transform and returns the payloads that survived `before_send`.
+      def process(event, options = nil)
+        processed = Sanitization.stringify_keys(event)
```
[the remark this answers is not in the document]
rafaeelaudibert: Not changing this. Request bodies are already capped at 4MB by the transport (`DEFAULT_MAX_REQUEST_BYTES`), tool results are server-generated, and the copy/sanitize/truncate passes are linear. Bounding before sanitization would also mean sanitizing truncated data.

lib/posthog/mcp/README.md
```diff
@@ -0,0 +1,116 @@
+# PostHog MCP analytics for Ruby
+
+> **Experimental.** `PostHog::MCP` is new and its API, options, and the captured `$mcp_*` event schema may change in a minor release. A one-line warning is logged when you require it. Please report issues at https://github.com/PostHog/posthog-ruby/issues.
```
marandaneto, line 3: lets remove all the code snippets and event/prop names from this file to https://posthog.com/docs/mcp-analytics so its a single source of truth

marandaneto:
@PostHog/mcp-analytics is there any demand for this (i've not seen any request for it)? its one more package to maintain so take that into consideration

lib/posthog/mcp/analytics.rb
```diff
@@ -0,0 +1,68 @@
...
+      def capture(event, properties = {})
+        unless event.is_a?(String) && !event.strip.empty?
+          raise ArgumentError, 'capture() requires an event name, e.g. analytics.capture("feedback_submitted")'
+        end
+
+        data = PostHog::MCP.tracking_data(@server)
+        return if data.nil?
+
+        Instrumentation.capture_event(data, {
+                                        'session_id' => data.session_id,
```
marandaneto: blocking: Scope custom events to the active request. `capture` reads server-global `data.session_id`, but the active request's session is prepared only after its tool returns. After Alice and then Bob initialize, a custom event emitted inside Alice's next tool call is attributed to Bob—even without concurrency. Retain request-local session and identity instead of using the last completed request's state. Reproduction: reproduced — the focused `review_instrumentation_regression_spec.rb` RSpec test received custom-event distinct ID `bob` while the corresponding automatic tool event correctly used `alice`.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,633 @@
...
+        stripped = safely({}) { strip_injected_arguments(arguments, owned) }
+        conversation_id, minted = safely([nil, false]) do
+          ConversationId.resolve(@options.enable_conversation_id, original_arguments, name, missing_name)
+        end
+        request = request_with_arguments(original_arguments)
+
+        if @options.report_missing && name == missing_name && !real_tool?(name)
+          result = Tools.result
+          safely { record_missing_capability(name, original_arguments, request) }
+          return result
```
marandaneto, line 150: blocking: Preserve MCP's lifecycle for virtual tool calls. This early return skips the underlying handler lambda, including its protocol-envelope validation and cancellation cleanup. MCP has already registered the stateful request's cancellation object, so every completed `get_more_tools` call retains an in-flight entry indefinitely; malformed modern-protocol requests also incorrectly succeed. The virtual response needs to run within the real request lifecycle. Reproduction: reproduced — focused MCP 1.5.0 transport tests retained three cancellation entries after three completed virtual calls and returned HTTP 200 for a malformed virtual call whose ordinary-tool control returned HTTP 400/-32602.

lib/posthog/mcp/sanitization.rb
```diff
@@ -0,0 +1,387 @@
...
+      def sanitize_response(response)
+        unless response.is_a?(Hash) || response.is_a?(Array) || response.is_a?(String)
+          return sanitize_captured_value(response)
+        end
+
+        sanitized = sanitize_captured_value(response)
+        return sanitized unless sanitized.is_a?(Hash)
+
+        result = sanitized.dup
+        if result['content'].is_a?(Array)
```
marandaneto: blocking: Redact binary prompt and resource response shapes too. Only top-level tool-result `content` arrays reach `sanitize_content_block`. Automatically captured prompt results instead use `messages[].content`, and resource reads use `contents[].blob`; these receive only generic string sanitization, so binary payloads below the 10KB heuristic threshold reach `$mcp_response` unchanged despite the documented binary-content redaction. Reproduction: reproduced — the focused `review_payloads_spec.rb` RSpec test found the original blob `c2Vuc2l0aXZl` in both `$mcp_prompt_get` and `$mcp_resource_read` payloads returned by the actual sink.

lib/posthog/mcp/sanitization.rb
```diff
@@ -0,0 +1,387 @@
...
+        return false unless prefix
+
+        payload = decode_percent(value[prefix[0].length..])
+        return false if payload.nil?
+
+        BASE64_DATA_URL_PAYLOAD_PATTERN.match?(payload.delete("\r\n"))
+      end
+
+      def decode_percent(value)
+        URI.decode_www_form_component(value)
```
marandaneto: blocking: Preserve literal plus signs when decoding base64 data URLs. `URI.decode_www_form_component` converts literal `+` into spaces, but `+` is a valid standard-base64 character in a data URL. Valid binary URLs then fail detection and can be transmitted unchanged. Use percent-decoding that preserves literal plus signs. Reproduction: reproduced — the focused `review_payloads_spec.rb` RSpec test passed `data:application/octet-stream;base64,` followed by `'++//' * 3000` through the sink and received the complete URL in `$mcp_parameters` instead of the binary placeholder.

lib/posthog/mcp/truncation.rb
```diff
@@ -0,0 +1,308 @@
...
+        ['client_name', MAX_METADATA_LENGTH],
+        ['client_version', MAX_METADATA_LENGTH],
+        ['error_type', MAX_METADATA_LENGTH],
+        ['client_user_agent', MAX_METADATA_LENGTH],
+        ['vendor_client', MAX_METADATA_LENGTH],
+        ['llm_model', MAX_METADATA_LENGTH]
+      ].freeze
+
+      NORMALIZED_FIELDS = %w[parameters response identify_actor_data error].freeze
+
```
marandaneto, line 57: blocking: Include custom properties in bounded normalization. User-controlled `properties` from custom events, `event_properties`, and `capture_tool_call` are excluded from normalization and progressive depth reduction. Large numeric arrays cannot be reduced by largest-string trimming, so oversized events survive this pipeline and are then discarded by the core batch despite the message-size guarantee. Reproduction: reproduced — the focused `review_payloads_spec.rb` RSpec test captured `properties: { rows: (1..10_000).to_a }`, produced a 49,288-byte message, and observed `PostHog::MessageBatch` retain zero messages instead of one.

lib/posthog/mcp/sanitization.rb
```diff
@@ -0,0 +1,387 @@
...
+      DIGIT_GROUP_PATTERN = /\d+/
+      PHONE_NANP_PATTERN = %r{(?<![\w+])(?:\+?1[ ./-]?)?(?:\(\d{3}\)[ ./-]?|\d{3}[ ./-])\d{3}[ ./-]\d{4}(?!\w)}
+      PHONE_INTL_PATTERN = %r{(?<!\w)\+\d{1,3}(?:[ ./()-]{0,2}\d){7,13}(?!\w)}
+
+      module_function
+
+      # Deep-copies a value with string keys so the pipeline can rely on one shape.
+      def stringify_keys(value)
+        case value
+        when Hash then value.to_h { |k, v| [k.to_s, stringify_keys(v)] }
```
marandaneto: blocking: Detect cycles before recursively copying telemetry. Ruby custom properties can contain a self-reference, but this traversal runs before the cycle-aware normalizer and recurses indefinitely. The resulting `SystemStackError` is not caught by the sink's `StandardError` rescue, so analytics crashes through into host code. This concerns Ruby custom properties, not oversized or cyclic JSON request bodies. Reproduction: reproduced — the focused `review_payloads_spec.rb` RSpec test supplied `properties['self'] = properties` and observed `SystemStackError` escaping the public `capture_tool_call` method before core-client dispatch.

lib/posthog/mcp/client.rb
```diff
@@ -0,0 +1,196 @@
...
+
+      # Pull the agent's intent off the injected `context` argument, strip it
+      # from the arguments, and flag the `get_more_tools` virtual tool.
+      #
+      # @return [PreparedToolCall]
+      def prepare_tool_call(name, args = nil)
+        raw_context = args.is_a?(Hash) ? (args[:context] || args['context']) : nil
+        intent = raw_context.is_a?(String) && !raw_context.strip.empty? ? raw_context.strip : nil
+        PreparedToolCall.new(
+          args: strip_context(args),
```
marandaneto: blocking: Preserve tool-owned context arguments in custom dispatchers. `prepare_tool_list` leaves a tool's existing `context` schema intact, but `prepare_tool_call` unconditionally removes that argument. A tool requiring its own context consequently receives incomplete arguments. Track which arguments the integration actually injected and strip only those. Reproduction: reproduced — the focused `review_custom_dispatch_spec.rb` RSpec test prepared a tool with its own required `context` field and expected `{ context: 'application data' }` after call preparation but received `{}`; an ownership-aware temporary control passed.

marandaneto (commented):
Advisory review of the complete PR patch at b72f5e0. These seven findings were reproduced with focused regression tests in disposable worktrees. The existing MCP and core-client suite passed (250 examples). Validation used Ruby 4.0.6, MCP 1.5.0, and available dependency versions rather than the complete locked bundle; temporary controls were not retained as production fixes.

marandaneto:
left a few comments but @PostHog/mcp-analytics should know more about the specifics

lib/posthog/mcp/analytics.rb
rafaeelaudibert: Fixed by e8f93cb.

Confirmed, and it did not need concurrency to trigger: `dispatch_tool_call` only resolved the session in `prepare_request` *after* the handler returned (the conversation anchor is only known then), so during Alice's tool body `data.session_id` still pointed at whoever finished last. Two layers in the fix:

1. **Prime before the tool body.** `Instrumentation#prime_session` resolves the transport/token session onto `data.session_id` before `yield`. Emits nothing; the post-call `prepare_request` still owns identify, lazy initialize and the conversation anchor.
2. **Request-local read.** The primed session is pinned to the in-flight `RequestScope` (the fiber-storage scope the transport already opens per HTTP request), and `Analytics#capture` reads it from there first, falling back to `data.session_id` only when there is no scope (stdio, single client). That covers overlapping requests on a threaded server too, which priming alone does not.

Specs in `http_transport_spec.rb` cover the sequential case you reproduced and an overlapping one (Alice's tool body lets Bob's whole request finish before it captures).

**Heads-up for `@posthog/mcp` (Node).** It has half of this. `captureToolCall` runs `prepareToolCallEvent` (which resolves the session and writes `data.sessionId`) *before* `execute(downstreamRequest)`, so the sequential Alice/Bob case is attributed correctly there. What it is missing is the second layer: `captureCustomEvent` in `packages/mcp/src/index.ts` reads `trackingData.sessionId`, one slot per server, and tool handlers are async. If Alice's handler awaits anything while Bob's request comes in and re-primes `data.sessionId`, Alice's `analytics.capture(...)` after the await is attributed to Bob. There is no `AsyncLocalStorage` (or equivalent request-local store) in the package today. The equivalent fix is to open an `AsyncLocalStorage` store per request in the request-handler patch, stash the primed `sessionId` in it, and have `captureCustomEvent` read from the store before falling back to `trackingData.sessionId`. Worth an issue on posthog-js.

lib/posthog/mcp/README.md
rafaeelaudibert: While this is experimental I thought it made sense to keep this here as source of truth. Eventually once the MCP Analytics starts owning this then they can probably own moving this. Are you ok with that?

rafaeelaudibert:
> @PostHog/mcp-analytics is there any demand for this (i've not seen any request for it)? its one more package to maintain so take that into consideration

@marandaneto I didn't hear any demand from them, but I have my side projects in Rails and I want to instrument them with MCP Analytics, so I built this for myself. I 100% believe we should not merge this if the team is not interested in maintaining it, because I don't plan on maintaining all this myself. I can get the same by doing a point implementation just for my server, too.

marandaneto:
> > @PostHog/mcp-analytics is there any demand for this (i've not seen any request for it)? its one more package to maintain so take that into consideration
> 
> @marandaneto I didn't hear any demand from them, but I have my side projects in Rails and I want to instrument them with MCP Analytics, so I built this for myself. I 100% believe we should not merge this if the team is not interested in maintaining it, because I don't plan on maintaining all this myself. I can get the same by doing a point implementation just for my server, too.

i mean all the work is done so i think we should merge and release it but i'd like to bring this up because i heard the same from the AIO team, they are too thin to bring support for more SDKs (or just maintain it)
its better to have the most important ones well maintained than a bunch not really

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by 6ccd422.

`instrument` now registers `get_more_tools` on the server with `define_tool` (skipped when the application already owns the name, which then behaves as an ordinary tool). The gem dispatches it like any other tool, so argument validation, envelope checks and the in-flight cancellation cleanup all run; the instrumentation only swaps the recorded event for `$mcp_missing_capability`. The manual `tools/list` append is gone since the gem lists it itself. Specs: stateful HTTP spec asserts no in-flight entries remain after three virtual calls, and a call without `context` now gets the gem's `Missing required arguments` error response while still recording the missing capability.

lib/posthog/mcp/sanitization.rb
rafaeelaudibert: Fixed by 6ca3657.

`decode_percent` uses `URI.decode_uri_component` (percent-only) with `URI::DEFAULT_PARSER.unescape` as the fallback on Ruby 3.0/3.1. Spec covers your `'++//' * 3000` payload and a percent-encoded plus.

lib/posthog/mcp/truncation.rb
rafaeelaudibert: Fixed by bbcc662.

`properties` is now part of `NORMALIZED_FIELDS`, so custom event properties, `event_properties` and `capture_tool_call` properties get breadth/depth normalization and take part in progressive reduction. Spec reproduces the 10k-row array and asserts the event lands under `MAX_EVENT_BYTES`.

lib/posthog/mcp/sanitization.rb
rafaeelaudibert: Fixed by 3615dea.

`stringify_keys` tracks visited containers by identity and emits the same `[Circular ~]` marker the normalizer uses (shared, non-cyclic references are still copied). `Client#capture_tool_call` with a self-referential `properties` hash no longer raises; covered in `client_spec` and `sanitization_spec`.

lib/posthog/mcp/client.rb
rafaeelaudibert: Fixed by 5f135d3.

`prepare_tool_call` takes an optional `input_schema:` (the tool's raw `inputSchema`, the same Hash handed to `prepare_tool_list`). When the schema declares `context`, the argument is left in `args`; otherwise it is the one we injected and it is stripped as before. Intent extraction is unchanged either way. Custom dispatchers are stateless across pods, so asking for the schema at call time is the only way to decide ownership reliably; remembering what `prepare_tool_list` injected would only be right on the pod that served the listing. Without the keyword the behaviour is unchanged. README updated, spec matches your reproduction.

lucasheriques:
i'll review this for the new spec soon, but as @marandaneto pointed, we on the MCP Analytics team do not have the capability to support this SDK for now.

i think we can merge this though - work is done as Manoel pointed, but maybe on the docs, we say it's not an official solution and we don't provide support for it yet.

lucasheriques (commented):
> [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

Replaced this long report with [nine concise inline comments and a short summary](https://github.com/PostHog/posthog-ruby/pull/269#pullrequestreview-5160070587).

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,649 @@
...
+      # schema does not declare.
+      def owned_params_for(name)
+        cached = @data.tool_owned_params[name]
+        return cached if cached
+
+        tools = @server.respond_to?(:tools) ? @server.tools : nil
+        tool = tools.is_a?(Hash) ? tools[name] : nil
+        schema = tool.respond_to?(:input_schema) ? tool.input_schema&.to_h : nil
+        owned = []
+        owned << 'context' if @options.context_enabled? && !SchemaMutation.declares_param?(schema, 'context')
```
lucasheriques, line 611: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P1] Preserve composed/reference schemas**

Before `tools/list`, a tool whose `allOf` schema requires application-owned `context` loses that argument and fails validation; the identical call succeeds after listing. Use one conservative injection/ownership guard, including `$ref`: injecting required `context` beside a reference to a closed object also makes discovery unsatisfiable. The standalone `prepare_tool_call` has the same composed-ownership problem.

lib/posthog/mcp/sanitization.rb
```diff
@@ -0,0 +1,407 @@
...
+      def sanitize_response(response)
+        unless response.is_a?(Hash) || response.is_a?(Array) || response.is_a?(String)
+          return sanitize_captured_value(response)
+        end
+
+        sanitized = sanitize_captured_value(response)
+        return sanitized unless sanitized.is_a?(Hash)
+
+        result = sanitized.dup
+        if result['content'].is_a?(Array)
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P1] Redact binary prompt/resource responses**

Only top-level `content` blocks are redacted. Real `prompts/get` and `resources/read` calls still capture short binary payloads from `messages[].content.data` and `contents[].blob` in `$mcp_response`. Extend shape-aware redaction to those response forms while preserving client-facing results. This reproduces the existing unresolved finding.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,649 @@
...
+        end
+
+        if error
+          event['is_error'] = true
+          event['error'] = Exceptions.capture_exception(error)
+        elsif !result.nil? && !no_response?(result)
+          event['response'] = result
+          if tool_result_error?(result)
+            event['is_error'] = true
+            event['error'] = Exceptions.capture_exception(Sanitization.stringify_keys(result))
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Extract errors before adding conversation metadata**

With `enable_conversation_id`, identical failures become `same failure {"conversation_id":"<fresh UUID>"}` in `$mcp_error_message`, fragmenting error grouping. Extract errors from the original result before appending the conversation handle. Current TS already does this.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,649 @@
...
+      # --- session / identity -----------------------------------------------
+
+      # Point the server-wide session at *this* request before the tool body runs,
+      # and pin it to the request scope. {Analytics#capture} reads the session from
+      # there, so a custom event emitted inside a tool belongs to its caller rather
+      # than to whichever request finished last (or is running concurrently). The
+      # full {#prepare_request} still runs after the call, because the conversation
+      # anchor is only known once the tool has returned. Emits nothing.
+      def prime_session
+        session_id, = Session.resolve(@data, mcp_session_id(@token), token: @token)
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Resolve identity and echoed conversation before capture**

On a first modern request with a valid echoed conversation ID, `analytics.capture` inside the tool uses an anonymous, temporary session; the automatic tool event uses the identified person and conversation session. Resolve and pin both identity and the already-known conversation anchor before invoking the tool.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,649 @@
...
+          end
+        end
+        stripped
+      end
+
+      def mutate_tool(tool)
+        return tool unless tool.is_a?(Hash)
+
+        name = fetch(tool, :name)
+        return tool if virtual_tool?(name)
```
lucasheriques, line 640: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Complete model capture for missing capabilities**

With `capture_model: true`, `get_more_tools` neither advertises `llm_model` nor records `$mcp_llm_model` when supplied. Its recorder and the standalone client's preparation/missing-capability helpers need the model support already present in current Python and TS.

lib/posthog/mcp/rack_middleware.rb
```diff
@@ -0,0 +1,111 @@
...
+            Log.debug(nil, "PostHog MCP session middleware: mint failed - #{e.message}")
+            nil
+          end
+          env[ENV_KEY] = SessionToken.decode(token) if token
+        end
+
+        status, headers, body = @app.call(env)
+        # Only a successful initialize gets the token, so a client cannot replay a
+        # session minted for a handshake the server rejected.
+        if token && success?(status) && headers.respond_to?(:key?) &&
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Check initialization success beyond HTTP status**

An HTTP 200 response containing a JSON-RPC `error` still gets a new `mcp-session-id`. The legacy transport contract associates that header with `InitializeResult`. Require a successful initialization result or an explicit success signal from the dispatcher before attaching the token.

lucasheriques (commented):
> [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

Recommend changes before merging. See the nine inline findings on MCP compatibility and SDK parity.

lib/posthog/mcp/identity.rb
```diff
@@ -0,0 +1,179 @@
...
+
+        _merged, changed = data.identified_sessions.merge!(session_id, identity)
+        return nil unless changed
+
+        Log.debug(data.options, "Identified session #{session_id}")
+        {
+          'session_id' => session_id,
+          'resource_name' => request_resource_name(request),
+          'event_type' => EventType::IDENTIFY,
+          'parameters' => { 'request' => request, 'extra' => captured_extra(extra) },
```
lucasheriques, line 120: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Remove raw intent from identify parameters**

With `identify` enabled, `context: "Look up jane@example.com"` is redacted on `$mcp_tool_call` but survives verbatim in `$identify` under `$mcp_parameters.request.params.arguments.context`. Use the captured-parameter builder here too. Python/TS share this privacy gap; parity alone would preserve it.

public_api_snapshot.txt
```diff
@@ -118,6 +118,145 @@ instance_method PostHog::Logging#logger()
...
+instance_method PostHog::MCP::Client#capture_missing_capability(context: ..., parameters: ..., protocol_version: ..., distinct_id: ..., session_id: ..., client_user_agent: ..., vendor_client: ..., set_properties: ..., groups: ..., properties: ..., timestamp: ...)
+instance_method PostHog::MCP::Client#capture_tool_call(tool_name, intent: ..., intent_source: ..., parameters: ..., response: ..., duration_ms: ..., is_error: ..., error: ..., error_type: ..., category: ..., tool_description: ..., protocol_version: ..., distinct_id: ..., session_id: ..., client_user_agent: ..., vendor_client: ..., set_properties: ..., groups: ..., properties: ..., timestamp: ..., llm_model: ..., llm_model_source: ...)
+instance_method PostHog::MCP::Client#capture_tools_list(tool_names: ..., parameters: ..., response: ..., duration_ms: ..., is_error: ..., error: ..., error_type: ..., protocol_version: ..., distinct_id: ..., session_id: ..., client_user_agent: ..., vendor_client: ..., set_properties: ..., groups: ..., properties: ..., timestamp: ...)
+instance_method PostHog::MCP::Client#prepare_tool_call(name, args = ...)
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Refresh the public API snapshot**

`prepare_tool_call` now accepts `input_schema:`, but this snapshot omits it. `bundle exec rake public_api:check` fails locally and in CI. Regenerate the snapshot after settling the API changes.

lib/posthog/mcp/schema_mutation.rb
```diff
@@ -0,0 +1,165 @@
...
+
+        if schema.nil? || (schema.respond_to?(:empty?) && schema.empty?)
+          schema = { type: 'object', properties: {},
+                     required: [] }
+        end
+        schema = deep_dup(schema)
+        properties_key = key_for(schema, :properties)
+        schema[properties_key] = {} unless schema[properties_key].is_a?(Hash)
+        additional_key = key_for(schema, :additionalProperties)
+        schema.delete(additional_key) if schema[additional_key] == false
```
lucasheriques: > [!NOTE]
> 🤖 Automated comment by **QA Swarm** — not written by a human

**[P2] Preserve additionalProperties: false**

For a strict tool declaring `x`, `{x: "yes", context: "why", typo: 123}` passes the advertised schema but dispatch rejects `/typo`. The injected property is already allowed through `properties`; keep `additionalProperties: false` so discovery matches validation. Python preserves it; TS shares this bug.

lib/posthog/mcp/sanitization.rb
rafaeelaudibert: Fixed by bf75806 — `sanitize_response` is now shape-aware for `messages[].content` (single block or array) and `contents[].blob`, so a sub-10KB blob is redacted there too. Covered end to end in `instrument_spec.rb` ("redacts binary prompt messages and resource blobs on the way out"), which asserts the original blob appears nowhere in the emitted payloads.

lib/posthog/mcp/sanitization.rb
rafaeelaudibert: Fixed by bf75806 — same fix as the blocking thread above: `messages[].content` and `contents[].blob` now go through the shape-aware redaction, with an end-to-end spec over real `prompts/get` and `resources/read` calls.

lib/posthog/mcp/request_scope.rb
```diff
@@ -0,0 +1,48 @@
...
+      FIBER_STORAGE = Fiber.respond_to?(:[]) && Fiber.respond_to?(:[]=)
+
+      module_function
+
+      # @return [Hash, nil] `{headers:, transport:, mint:, session_id:}` for the in-flight HTTP request.
+      #   `session_id` is the `$session_id` {Instrumentation} settled on before running the
+      #   tool body, so a custom event captured inside the tool is attributed to this
+      #   request even while another request on the same server is in flight.
+      def current
+        FIBER_STORAGE ? Fiber[KEY] : Thread.current[KEY]
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed by bf75806 — fails closed instead of falling back. Once a request has arrived over HTTP, a capture with no request scope gets a standalone `$session_id` (and a one-time warning) rather than the server-wide session; over stdio, where a server only ever talks to one client, the fallback stays. Ruby 3.2+ still inherits the scope into spawned threads/fibers via `Fiber[]`.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by bf75806 — one guard, `SchemaMutation.injectable?` (not composed, no `$ref`), is now used by `add_parameter`, `Instrumentation#owned_params_for` and `Client#prepare_tool_call`. A composed or referenced schema is neither injected into nor stripped from, so a call before the first `tools/list` behaves exactly like one after it.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by bf75806 — the error is now extracted from the result the tool returned, before the conversation handle is appended, so identical failures keep an identical `$mcp_error_message`. The handle still reaches the agent on the delivered result.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by bf75806 — session and identity are now settled before the tool body runs, anchored on the echoed `conversation_id` when the agent sent one. Non-HTTP transports open a request scope too, so the pin works over stdio as well, and `identify` still runs once per session per request.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by bf75806 — `get_more_tools` advertises `llm_model` under `capture_model` and records `$mcp_llm_model`/`$mcp_llm_model_source`, in the instrumented path and in `Client#prepare_tool_list` / `#capture_missing_capability`. `conversation_id` is deliberately left off the virtual tool: it reports a gap in the tool list rather than taking part in a tool conversation.

lib/posthog/mcp/rack_middleware.rb
rafaeelaudibert: Fixed by bf75806 — the middleware now requires an `InitializeResult` before attaching `Mcp-Session-Id`, so a JSON-RPC error on a 200 mints nothing. A body that cannot be read without consuming a stream (SSE) keeps the status-only behaviour, which is documented on the method.

public_api_snapshot.txt
rafaeelaudibert: Fixed by bf75806 — regenerated after the API changes settled; `bundle exec rake public_api:check` passes.

lib/posthog/mcp/schema_mutation.rb
rafaeelaudibert: Fixed by bf75806 — `additionalProperties: false` is preserved; the injected name is listed under `properties`, so it is still accepted, and the spec that asserted the removal was inverted.

lib/posthog/mcp/identity.rb
rafaeelaudibert: Not changing this one: we want the same behaviour across the SDKs here, so this stays as it is rather than diverging in Ruby alone. Worth raising as a cross-SDK issue if we want the captured-parameter builder used for `$identify` everywhere.

rafaeelaudibert:
Pushed bf75806, which addresses the outstanding review threads. Summary of what changed and how:

| Thread | Fix |
|---|---|
| Binary prompt/resource responses not redacted (blocking, + the QA Swarm duplicate) | `sanitize_response` is shape-aware for `prompts/get` `messages[].content` (single block or array) and `resources/read` `contents[].blob`, not just tool-result `content`. End-to-end spec asserts the blob appears nowhere in the emitted payloads. |
| Request scope lost in child fibers/threads on Ruby 3.0/3.1 | Fails closed: once a request has arrived over HTTP, a capture with no request scope gets a standalone `$session_id` plus a one-time warning instead of the server-wide session. Over stdio the fallback stays. |
| Composed/`$ref` schemas lose a tool-owned `context` | One guard, `SchemaMutation.injectable?`, used by `add_parameter`, `Instrumentation#owned_params_for` and `Client#prepare_tool_call`: a composed or referenced schema is neither injected into nor stripped from. |
| Conversation handle fragments error grouping | The error is extracted from the result the tool returned, before the handle is appended. The handle still reaches the agent on the delivered result. |
| In-tool events use an anonymous, wrong session | Session and identity settle before the tool body runs, anchored on an echoed `conversation_id`. Non-HTTP transports now open a request scope too, and `identify` still runs once per session per request. |
| `get_more_tools` missing model capture | The virtual tool advertises `llm_model` under `capture_model` and records `$mcp_llm_model`; `Client#prepare_tool_list` and `#capture_missing_capability` gained the same support. `conversation_id` is deliberately left off the virtual tool. |
| Token minted for a failed initialize | `RackMiddleware` requires an `InitializeResult` before attaching `Mcp-Session-Id`; a body that cannot be sniffed without consuming a stream (SSE) keeps the status-only behaviour. |
| `additionalProperties: false` dropped | Preserved — the injected name is listed under `properties`, so it is still accepted. The spec that asserted the removal was inverted. |
| Public API snapshot stale | Regenerated; `rake public_api:check` passes. |

Two threads intentionally unchanged:

- **Raw intent in `$identify` parameters** — keeping the behaviour consistent across the SDKs rather than diverging in Ruby alone; better raised as a cross-SDK issue.
- **Moving the README content to the docs site** — replied in-thread, waiting on @marandaneto.

`bundle exec rspec` (839 examples) and `rubocop` are clean.

lib/posthog/mcp/README.md
marandaneto: well fine by me but mcp in general is experimental https://github.com/PostHog/posthog-js/blob/f998ee9c0010d94f3fa0f33cccfd57f55e792e7e/packages/mcp/package.json#L3
and it lives in posthog.com
same for python, so i dont see a reason not to

lib/posthog/mcp/rack_middleware.rb
```diff
@@ -0,0 +1,158 @@
...
+      # Reads the request body (bounded) and makes it re-readable for the app.
+      def read_body(env)
+        input = env['rack.input']
+        return '' unless input.respond_to?(:read)
+
+        body = input.read(MAX_SNIFF_BODY + 1) || ''
+        if input.respond_to?(:rewind)
+          input.rewind
+        else
+          rest = input.read || ''
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed by f6bf9f0 — the middleware no longer reads the request body at all, so there is nothing to buffer. It publishes the request headers to `RequestScope` as an HTTP request and carries back the `Mcp-Session-Id` an instrumented server below it minted once the handshake produced an `InitializeResult`; a hand-rolled dispatcher, which has already parsed the body itself, mints through `env['posthog_mcp.mint']`. The response-body sniff is gone with it, so the SSE caveat and the 200-with-a-JSON-RPC-error heuristic both disappear.

lib/posthog/mcp/README.md
rafaeelaudibert: Done in f6bf9f0 — you're right, and posthog.com is now the single source of truth. `lib/posthog/mcp/README.md` keeps no code snippets or event/property names: just what the integration is, a link to https://posthog.com/docs/mcp-analytics, the handful of genuinely Ruby-specific notes (per-event `$lib`, the tighter 32KB truncation budget, request scope needing Ruby 3.2+ to be inherited, composed schemas left untouched, stdio logging), and a file map for anyone reading the code.

While I was there I also took @lucasheriques' point about support: the README banner, the require-time warning, the changeset and the example now say the integration is experimental **and not officially supported**, rather than implying it's supported via issues. Same on the docs side — https://github.com/PostHog/posthog.com/pull/20022 now says the MCP analytics team doesn't maintain it and points at TypeScript or Python for a supported SDK.

marandaneto:
lgtm
i'll let @PostHog/mcp-analytics hit the approval button if they are fine with it

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,703 @@
...
+        end
+
+        # Enrich an event with session/identity/server metadata and hand it to
+        # the sink.
+        def capture_event(data, input)
+          sink = data.sink
+          return nil if sink.nil?
+
+          session_id = input['session_id'] || data.session_id
+          actor = session_id ? data.identified_sessions.get(session_id) : nil
```
lucasheriques, line 53: I found a race in identity attribution. `capture_event` reads the shared identity cache after the tool finishes. Another call in the same session can change it first. In a two-thread test, Alice's call waited while Bob's finished, and Alice's `$mcp_tool_call` used Bob's `distinct_id`.

The JavaScript SDK keeps the resolved identity with each request. Please keep that request identity for automatic and custom events, and add a concurrent regression test.

lib/posthog/mcp/client.rb
```diff
@@ -0,0 +1,226 @@
...
+      # left in `args`: only an injected `context` is stripped. A composed
+      # (oneOf/allOf/anyOf) or `$ref` schema is never injected into, so its
+      # `context` is the tool's own and is left alone too. Without the schema
+      # there is no way to tell the two apart, so `context` is always stripped.
+      #
+      # @param name [String] tool name
+      # @param args [Hash, nil] the call's arguments
+      # @param input_schema [Hash, nil] the tool's raw `inputSchema`
+      # @return [PreparedToolCall]
+      def prepare_tool_call(name, args = nil, input_schema: nil)
```
lucasheriques, line 163: The custom-dispatcher helpers do not round-trip `llm_model` yet. `prepare_tool_list(..., capture_model: true)` adds it as a required argument, but `prepare_tool_call` leaves it in `args` and does not return the model or its source. I passed those args to a strict Ruby tool and got `ArgumentError: unknown keyword: :llm_model`.

JavaScript and Python remove SDK-owned model fields and return the model metadata. Please add the same handling here, with a round-trip test.

lucasheriques (commented):
I reviewed the Ruby MCP integration against the current JavaScript and Python SDKs and ran the Ruby test suite. I left five inline findings for discussion.

lib/posthog/mcp/sink.rb
```diff
@@ -0,0 +1,86 @@
...
+          processed = Truncation.truncate_event(processed)
+        rescue StandardError => e
+          Log.debug(options, "Failed to truncate event: #{e.message}")
+          return nil
+        end
+        processed['id'] = Ids.new_prefixed_id('evt') unless processed['id'].is_a?(String) && !processed['id'].empty?
+
+        autocapture = options.nil? || options.enable_exception_autocapture
+        payloads = EventBuilder.build(processed, enable_exception_autocapture: autocapture)
+        apply_before_send(payloads, options)
```
lucasheriques, line 52: An expanding `before_send` hook can cause a silent event drop. The MCP size check runs before the hook, while `MessageBatch` applies Ruby's 32 KB limit afterward. A hook that added 40 KB made the sink return one payload, but the transport sent zero requests.

Please enforce the final size limit after `before_send` and test this case.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,703 @@
...
+
+      def dispatch_tool_call
+        name = fetch(@params, :name)
+        arguments = fetch(@params, :arguments)
+        arguments = nil unless arguments.is_a?(Hash)
+        original_arguments = arguments&.dup
+        missing_name = Tools.missing_capability_tool_name(@options)
+        owned = safely([]) { owned_params_for(name) }
+        stripped = safely({}) { strip_injected_arguments(arguments, owned) }
+        conversation_id, minted = safely([nil, false]) do
```
lucasheriques, line 137: The ownership check is in place, but intent and conversation resolution still read application-owned fields from `original_arguments`. I saw a tool-owned `context` become `$mcp_intent`. With two users sharing a tool-owned UUIDv7 `conversation_id`, Bob's event also included Alice's email in `$set`.

The JavaScript SDK consumes these values only when analytics owns them. Please use the ownership result for both fields and add tests for tools that declare them.

lib/posthog/mcp/instrumentation.rb
```diff
@@ -0,0 +1,703 @@
...
+          @identified_in_request[session_id] = true
+          identify_event = Identity.handle_identify(@data, session_id, request, extra)
+          self.class.capture_event(@data, identify_event) if identify_event
+        end
+        maybe_emit_initialize(session_id, request) unless skip_initialize
+        session_id
+      end
+
+      def maybe_emit_initialize(session_id, request)
+        return if @data.session_initialized?(session_id)
```
lucasheriques, line 438: The initialize check and mark are locked separately, but the full step is not atomic. Two threads can pass the check before either marks the session. I reproduced two `$mcp_initialize` events for one session.

Please make check-and-mark one atomic operation and add a concurrent test.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by a21946f.

`Identity.identify_for_request` now hands the resolved actor back rather than leaving it in the session-keyed cache to be read again once the handler returns, and the actor is pinned for the rest of the request. `RequestScope` carries it next to the session id, so a custom event captured from inside a tool body follows the same identity instead of re-reading shared state. `capture_event` consults the cache only for a caller with no request identity to offer — an `UNRESOLVED_ACTOR` sentinel keeps that distinct from a request that resolved to nobody, so a lost scope fails closed on identity the same way it already does on session.

Regression test parks one request inside its tool body on a queue, lets a second request identify and finish on the same session, then releases the first, so it is deterministic rather than timing-dependent. Against the old code it fails exactly as you described (`expected: "alice", got: "bob"`), and reverting only the `RequestScope` half fails the custom-event assertion on its own.

lib/posthog/mcp/client.rb
rafaeelaudibert: Fixed by 7d85d80.

`prepare_tool_call` now strips an injected `llm_model` and returns `llm_model`/`llm_model_source`, ready to hand straight to `capture_tool_call`. Ownership is decided per argument by the same schema check `context` already used, so a tool that declares either field itself keeps it in `args` and it is not reported.

Round-trip test covers both directions and splats the prepared args into a lambda with required keywords, which is your `ArgumentError` repro. `PreparedToolCall` gained two members, so `public_api_snapshot.txt` is updated.

One extra thing I found while in there: `prepare_tool_call` was still reading a tool-declared `context` as `intent`, even though 5f135d3 had stopped stripping it from `args`. Same defect as your other comment, so I fixed it here too — two specs asserted the old behaviour and now assert the new.

lib/posthog/mcp/sink.rb
rafaeelaudibert: Fixed by 7d85d80.

The byte budget is re-applied to whatever `before_send` returns, via `Truncation.truncate_payload`. I went with re-truncating rather than dropping: a hook that enriches an event should not cost you the event.

Small correction to the report — the drop is not quite silent, `MessageBatch` logs `a message exceeded the maximum allowed size` at error level. The event was still lost, though, and `MAX_EVENT_BYTES` exists precisely to stay under that limit, so the guarantee was being bypassed.

Test drives a hook that adds 40 KB and asserts both events still ship and that each is within `Defaults::Message::MAX_BYTES`.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by 7d85d80.

`dispatch_tool_call` now resolves both the conversation id and the intent from the arguments this layer stripped, never `original_arguments`. Model capture already read the stripped value; the other two now agree with it, so ownership is decided in one place for all three.

Tests cover a tool that declares `context` (no `$mcp_intent`), a composed schema (same, listed or not), and a tool that declares `conversation_id` — two calls sharing one application-owned handle now mint separate ids and are not stitched onto one session, which is what was bleeding `$set` across users.

lib/posthog/mcp/instrumentation.rb
rafaeelaudibert: Fixed by 7d85d80.

Check and mark are one critical section now, in `TrackingData#claim_session_initialized`: only the caller that inserts the key gets `true` and emits. The explicit `initialize` path still marks unconditionally, since that one should always emit.

Concurrent test runs 12 threads released together off a queue across 50 sessions and asserts exactly one claim each. I checked it against a non-atomic implementation of the same method: it fails 8 out of 8 runs.

lib/posthog/mcp/sanitization.rb
```diff
@@ -0,0 +1,436 @@
...
+
+      def sanitize_string(value)
+        return BINARY_REDACTED_VALUE if binary_like?(value)
+
+        redact_secret_tokens(value.gsub(POSTHOG_TOKEN_PATTERN, REDACTED_VALUE))
+      end
+
+      # Redact credential-looking words, leaving surrounding text intact.
+      def redact_secret_tokens(value)
+        return (SecretDetection.secret?(value) ? REDACTED_VALUE : value) unless value.include?(' ')
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed by 90d38c7, though not quite as reported — the mechanism holds, the conclusion does not.

**PEM keys were not being forwarded unredacted.** The split you flag is what saved them: each base64 body line becomes its own whitespace-free word and trips the entropy check on its own. Running real keys through `sanitize_string`, 0 of 49 body chunks leaked for RSA 2048 (5 keys), 0 of 25 for RSA 1024, 0 of 5 for EC, 0 of 50 for PKCS#8. Output is `-----BEGIN RSA PRIVATE KEY----- [redacted] [redacted] … -----END RSA PRIVATE KEY-----`; 1675 chars in, 337 out. Only the BEGIN/END markers survive.

**The marker really was unreachable, and worse than you describe** — not just for multi-line PEMs, but from both call sites, always. `PEM_PRIVATE_KEY_MARKER` contains a space, `sanitization.rb:116` is only reached when the value has no space, and `:118` tests single words. No input could ever reach it. Dead code, not a hole.

**Chasing it did find a real gap, with a different cause.** `path_or_url?` reads a body line with two lowercase `/`-separated segments as a file path, which suppresses the entropy check:

```
fnjv/87b1pybnOeVFm/c/oO4xX5sBmCi6GCRnw4WOAvKbBEtMobMdZVh6gSm5ni3
  entropy_path=false  path_or_url=true  -> survived
```

3 of 2600 real key lines (~0.12%), so roughly one key in twenty-five leaked a 64-char fragment. Not key recovery, but key material reaching analytics.

**What landed.** I took the block rather than the whole value: `-----BEGIN … PRIVATE KEY-----` blocks are matched and redacted whole before anything is split into words, so no part of a key rides on entropy scoring. Your version would redact the entire value, losing all surrounding prose, which cuts against the "redact credential-looking words, leaving surrounding text intact" design. Matching is non-greedy and terminates at end of string, so multiple blocks are handled separately and a truncated block still loses its body. Covers RSA/EC/ENCRYPTED/OPENSSH/PKCS#8 and the PGP `BLOCK` spelling. The unreachable marker check is deleted with a note, since dead code that looks protective is what produced this report.

Verified at 0 leaked chunks across 400 keys and every variant, with the path-looking line above as a regression test. No backtracking blowup: 540 KB plain text 17 ms, `BEGIN` with no `END` over 500 KB 13 ms.

**Left alone deliberately:** `path_or_url?` still shadows any `/`-bearing base64 secret, not just PEMs. The obvious fix — requiring path-like segments to cover most of the value — reclassifies 80 of the 190 real paths in this repo, because `PATH_WORD_RE` admits neither digits nor underscores (`backoff_policy.rb` is not a path word). Redacting real file paths is the worse failure, so that needs its own change starting with the character class.

lib/posthog/mcp/exceptions.rb
```diff
@@ -0,0 +1,112 @@
...
+      def capture_exception(error)
+        return from_message(call_tool_result_message(error)) if call_tool_result?(error)
+        return from_exception(error) if error.is_a?(Exception)
+        return from_message(error) if error.is_a?(String)
+
+        from_message(safe_to_s(error))
+      end
+
+      def from_exception(error)
+        list = PostHog::ExceptionCapture.build_exception_list(error) || []
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed by 6f6d840. This one reproduces exactly as reported, so thanks — confirmed before fixing:

```
frame: secret_tool.rb:14 in_app=true
  context_line: "      raise ArgumentError, 'boom'"
  pre_context:
    "  STRIPE_KEY = 'sk_live_…'"
    "  DB_PASSWORD = 'hunter2-correct-horse-battery-staple'"
```

`add_context_lines` attaches five lines either side of the raise for every `in_app` frame, `sanitize_event` routes `error` only through `sanitize_exception_values`, and that reached nothing but each entry's `value`. Autocapture is on by default.

Worth naming why this is worth closing here rather than being ordinary error-tracking behaviour, since the core SDK does the same for every Ruby app: the trigger is remote and untrusted — a caller can make a tool fail on demand and harvest source — and this layer already sanitizes exception values, so it advertised redaction inside exceptions and then skipped the bulk of the payload. That inconsistency was the real defect.

**Sanitized rather than stripped.** `pre_context`, `context_line` and `post_context` now get the same redaction as every other captured string. Stripping would guarantee more, but it would leave MCP errors harder to debug than any other error this SDK reports, and context lines are the most useful part of a trace.

One implementation note: replacement happens per non-space run rather than through the usual split-and-rejoin, because that path rejoins on single spaces and would flatten indentation — `"      raise ArgumentError"` came back as `"raise ArgumentError"`. Shape is now preserved:

```
pre:  "  API_KEY = [redacted]"
pre:  "  class << self"
pre:  "    def call(**)"
line: "      raise ArgumentError, 'boom'"
```

**Be clear about the limit**, since "sanitize" can read as stronger than it is: this is the limit of the detector, not the plumbing. Recognisable credential formats and high-entropy strings go; a passphrase-style password does not. Measured on the repro, `sk_live_…`, `ghp_…` and `AKIA…` are all redacted, `hunter2-correct-horse-battery-staple` survives. That is the same guarantee the rest of the pipeline offers — no stronger — and anyone needing an absolute one should set `enable_exception_autocapture: false`.

Also worth recording: the first push of this fix was blocked by GitHub push protection, because the spec used Stripe's documentation example key as a fixture. Swapped for non-scannable fixtures.

lib/posthog/mcp/sanitization.rb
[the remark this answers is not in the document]
rafaeelaudibert: Thanks for verifying. Noting the correction for the record, since the original report read as a live bypass: PEM keys were never forwarded unredacted. The split caught each base64 body line on entropy, so 0 of 49 body chunks leaked for RSA 2048. What was actually wrong was that the marker check was unreachable from both call sites, and that the entropy path it stood in for has a `path_or_url?` hole that leaked roughly one body line in a thousand. Both are closed now.

lucasheriques (approved):
Thanks for all the careful work here, Rafa. I reviewed the fixes for the five findings and the later redaction issue, and I reran the MCP test suite. Everything looks good from my side. Nice work closing the concurrency and security gaps with focused regression tests.

